PR #109, opened 2019-02-28
Calculating a new target rate with every rebase using data from CPI Oracle

nithinkrishna:
- The policy now has no notion of volume.
- Using a cpi Oracle to compute the new `targetRate` on every rebase.

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -71,7 +71,7 @@ contract UFragmentsPolicy is Ownable {
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
     // CPI-U value July 10th 1983 100, DECIMALS Fixed point number
-    uint256 private constant BASE_CPI = 100 * 10**(DECIMALS);
+    uint256 private constant BASE_CPI = 100 * (10**DECIMALS);
 
     /**
      * @notice Anyone can call this function to initiate a new rebase operation, provided more than
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -8,8 +8,8 @@ import "./lib/UInt256Lib.sol";
 import "./UFragments.sol";
 
 
-interface IMarketOracle {
-    function getPriceAnd24HourVolume() external returns (uint256, uint256);
+interface IOracle {
+    function getData() external returns (uint256, bool);
 }
 
 
@@ -30,19 +30,24 @@ contract UFragmentsPolicy is Ownable {
     event LogRebase(
         uint256 indexed epoch,
         uint256 exchangeRate,
-        uint256 volume24hrs,
+        uint256 cpi,
         int256 requestedSupplyAdjustment
     );
 
     UFragments public uFrags;
-    IMarketOracle public marketOracle;
 
-    // If the current exchange rate is within this absolute distance from the target, no supply
+    // Cpi oracle provides the current CPI-U, seasonally adjusted data,
+    // as a 18 decimal fixed point number
+    IOracle public cpiOracle;
+
+    // Market oracle provides the token/USD exchange rate as a 18 decimal fixed point number
+    // (eg) An oracle value of 1.5e18 it would mean 1 Ample is trading for $1.50.
+    IOracle public marketOracle;
+
+    // If the current exchange rate is within this fractional distance from the target, no supply
     // update is performed. Fixed point number--same format as the rate.
+    // (ie) abs(rate - targetRate) / targetRate < deviationThreshold, then no supply change.
     uint256 public deviationThreshold;
-
-    // The 24hr market volume must be at least this value before any supply adjustments occur.
-    uint256 public minimumVolume;
 
     // The rebase lag parameter, used to dampen the applied supply adjustment by 1 / rebaseLag
     // Check setRebaseLag comments for more details.
@@ -57,23 +62,23 @@ contract UFragmentsPolicy is Ownable {
     // The number of rebase cycles since inception
     uint256 public epoch;
 
-    uint256 private constant RATE_DECIMALS = 18;
-
-    uint256 private constant TARGET_RATE = 1 * 10**RATE_DECIMALS;
-
-    int256 private constant TARGET_RATE_SIGNED = int256(TARGET_RATE);
+    uint256 private constant DECIMALS = 18;
 
     // Due to the expression in computeSupplyDelta(), MAX_RATE * MAX_SUPPLY must fit into an int256.
     // Both are 18 decimals fixed point numbers.
-    uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
+    uint256 private constant MAX_RATE = 10**6 * 10**DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
+
+    // CPI-U value July 10th 1983 100, DECIMALS Fixed point number
+    uint256 private constant BASE_CPI = 100 * 10**(DECIMALS);
 
     /**
      * @notice Anyone can call this function to initiate a new rebase operation, provided more than
      *         the minimum time period has elapsed.
      * @dev The supply adjustment equals (_totalSupply * DeviationFromTargetRate) / rebaseLag
-     *      Where DeviationFromTargetRate is (MarketOracleRate - TARGET_RATE) / TARGET_RATE
+     *      Where DeviationFromTargetRate is (MarketOracleRate - targetRate) / targetRate
+     *      and targetRate is CpiOracleRate / BASE_CPI
      */
     function rebase() external {
         // This comparison also ensures there is no reentrancy.
@@ -81,14 +86,24 @@ contract UFragmentsPolicy is Ownable {
         lastRebaseTimestampSec = now;
         epoch = epoch.add(1);
 
+        uint256 cpi;
+        bool cpiValid;
+        (cpi, cpiValid) = cpiOracle.getData();
+        require(cpiValid);
+
+        uint256 targetRate = cpi.mul(10 ** DECIMALS).div(BASE_CPI);
+
         uint256 exchangeRate;
-        uint256 volume;
-        (exchangeRate, volume) = marketOracle.getPriceAnd24HourVolume();
+        bool rateValid;
+        (exchangeRate, rateValid) = marketOracle.getData();
+        require(rateValid);
+
         if (exchangeRate > MAX_RATE) {
             exchangeRate = MAX_RATE;
         }
 
-        int256 supplyDelta = computeSupplyDelta(exchangeRate, volume);
+        int256 supplyDelta = computeSupplyDelta(exchangeRate, targetRate);
+
         // Apply the Dampening factor.
         supplyDelta = supplyDelta.div(rebaseLag.toInt256Safe());
 
@@ -98,14 +113,25 @@ contract UFragmentsPolicy is Ownable {
 
         uint256 supplyAfterRebase = uFrags.rebase(epoch, supplyDelta);
         assert(supplyAfterRebase <= MAX_SUPPLY);
-        emit LogRebase(epoch, exchangeRate, volume, supplyDelta);
+        emit LogRebase(epoch, exchangeRate, cpi, supplyDelta);
+    }
+
+    /**
+     * @notice Sets the reference to the CPI oracle.
+     * @param cpiOracle_ The address of the cpi oracle contract.
+     */
+    function setCpiOracle(IOracle cpiOracle_)
+        external
+        onlyOwner
+    {
+        cpiOracle = cpiOracle_;
     }
 
     /**
      * @notice Sets the reference to the market oracle.
      * @param marketOracle_ The address of the market oracle contract.
      */
-    function setMarketOracle(IMarketOracle marketOracle_)
+    function setMarketOracle(IOracle marketOracle_)
         external
         onlyOwner
     {
@@ -113,28 +139,16 @@ contract UFragmentsPolicy is Ownable {
     }
 
     /**
-     * @notice Sets the deviation threshold. If the exchange rate given by the market
-     *         oracle is within this absolute distance from the target, then no supply
-     *         modifications are made. RATE_DECIMALS fixed point number.
-     * @param deviationThreshold_ The new exchange rate threshold.
+     * @notice Sets the deviation threshold fraction. If the exchange rate given by the market
+     *         oracle is within this fractional distance from the targetRate, then no supply
+     *         modifications are made. DECIMALS fixed point number.
+     * @param deviationThreshold_ The new exchange rate threshold fraction.
      */
     function setDeviationThreshold(uint256 deviationThreshold_)
         external
         onlyOwner
     {
         deviationThreshold = deviationThreshold_;
-    }
-
-    /**
-     * @notice Sets the minimum volume. During rebase, the volume must be at least this value before
-     * any supply adjustment is made.
-     * @param minimumVolume_ The new minimum volume, measured in 24hr market Token Volume.
-     */
-    function setMinimumVolume(uint256 minimumVolume_)
-        external
-        onlyOwner
-    {
-        minimumVolume = minimumVolume_;
     }
 
     /**
@@ -177,8 +191,9 @@ contract UFragmentsPolicy is Ownable {
     {
         Ownable.initialize(owner);
 
-        deviationThreshold = (5 * TARGET_RATE) / 100;  // 5% of target
-        minimumVolume = 1;
+        // deviationThreshold = 0.05e18 = 5e16
+        deviationThreshold = 5 * 10 ** (DECIMALS-2);
+
         rebaseLag = 30;
         minRebaseTimeIntervalSec = 1 days;
         lastRebaseTimestampSec = 0;
@@ -188,46 +203,40 @@ contract UFragmentsPolicy is Ownable {
     }
 
     /**
-     * @return Computes the total supply adjustment in response to the exchange rate.
+     * @return Computes the total supply adjustment in response to the exchange rate
+     *         and the targetRate.
      */
-    function computeSupplyDelta(uint256 rate, uint256 volume)
+    function computeSupplyDelta(uint256 rate, uint256 targetRate)
         private
         view
         returns (int256)
     {
-        if (withinDeviationThreshold(rate) || !enoughVolume(volume)) {
+        if (withinDeviationThreshold(rate, targetRate)) {
             return 0;
         }
 
-        // (totalSupply * (rate - target)) / target
-        return uFrags.totalSupply().toInt256Safe().mul(
-            rate.toInt256Safe().sub(TARGET_RATE_SIGNED)
-        ).div(TARGET_RATE_SIGNED);
+        // supplyDelta = totalSupply * (rate - targetRate) / targetRate
+        int256 targetRateSigned = targetRate.toInt256Safe();
+        return uFrags.totalSupply().toInt256Safe()
+            .mul(rate.toInt256Safe().sub(targetRateSigned))
+            .div(targetRateSigned);
     }
 
     /**
      * @param rate The current exchange rate, an 18 decimal fixed point number.
+     * @param targetRate The target exchange rate, an 18 decimal fixed point number.
      * @return If the rate is within the deviation threshold from the target rate, returns true.
      *         Otherwise, returns false.
      */
-    function withinDeviationThreshold(uint256 rate)
+    function withinDeviationThreshold(uint256 rate, uint256 targetRate)
         private
         view
         returns (bool)
     {
-        return (rate >= TARGET_RATE && rate.sub(TARGET_RATE) < deviationThreshold)
-            || (rate < TARGET_RATE && TARGET_RATE.sub(rate) < deviationThreshold);
-    }
+        uint256 absoluteDeviationThreshold = targetRate.mul(deviationThreshold)
+            .div(10 ** DECIMALS);
 
-    /**
-     * @param volume Total trade volume of the last reported 24 hours in Token volume.
-     * return True, if the volume meets requirements for a supply adjustment. False otherwise.
-     */
-    function enoughVolume(uint256 volume)
-        private
-        view
-        returns (bool)
-    {
-        return volume >= minimumVolume;
+        return (rate >= targetRate && rate.sub(targetRate) < absoluteDeviationThreshold)
+            || (rate < targetRate && targetRate.sub(rate) < absoluteDeviationThreshold);
     }
 }
--- a/contracts/mocks/MockOracle.sol
+++ b/contracts/mocks/MockOracle.sol
@@ -3,32 +3,37 @@ pragma solidity 0.4.24;
 import "./Mock.sol";
 
 
-contract MockMarketOracle is Mock {
-    uint128 private _exchangeRate;
-    uint128 private _volume;
+contract MockOracle is Mock {
+    bool private _validity = true;
+    uint256 private _data;
+    string public name;
+
+    constructor(string name_) public {
+        name = name_;
+    }
 
     // Mock methods
-    function getPriceAnd24HourVolume()
+    function getData()
         external
-        returns (uint128, uint128)
+        returns (uint256, bool)
     {
-        emit FunctionCalled("MarketOracle:getPriceAnd24HourVolume", msg.sender);
+        emit FunctionCalled(name, "getData", msg.sender);
         uint256[] memory uintVals = new uint256[](0);
         int256[] memory intVals = new int256[](0);
         emit FunctionArguments(uintVals, intVals);
-        return (_exchangeRate, _volume);
+        return (_data, _validity);
     }
 
     // Methods to mock data on the chain
-    function storeRate(uint128 exchangeRate)
+    function storeData(uint256 data)
         public
     {
-        _exchangeRate = exchangeRate;
+        _data = data;
     }
 
-    function storeVolume(uint128 volume)
+    function storeValidity(bool validity)
         public
     {
-        _volume = volume;
+        _validity = validity;
     }
 }
--- a/test/unit/UFragmentsPolicy.js
+++ b/test/unit/UFragmentsPolicy.js
@@ -1,6 +1,6 @@
 const UFragmentsPolicy = artifacts.require('UFragmentsPolicy.sol');
 const MockUFragments = artifacts.require('MockUFragments.sol');
-const MockMarketOracle = artifacts.require('MockMarketOracle.sol');
+const MockOracle = artifacts.require('MockOracle.sol');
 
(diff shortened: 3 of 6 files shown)


contracts/UFragments.sol
```diff
@@ -72,6 +72,8 @@ contract UFragments is ERC20Detailed, Ownable {
 
     uint256 private constant DECIMALS = 9;
     uint256 private constant MAX_UINT256 = ~uint256(0);
+
+    // TODO: This needs to be updated before launch
```
brandoniles: I'm not worried about forgetting this, and I'd rather avoid code churn between now and launch as much as possible.

contracts/UFragmentsPolicy.sol
```diff
@@ -8,8 +8,8 @@ import "./lib/UInt256Lib.sol";
 import "./UFragments.sol";
 
 
-interface IMarketOracle {
-    function getPriceAnd24HourVolume() external returns (uint256, uint256);
+interface IOracle {
+    function getData() external returns (bool, uint256);
```
brandoniles: Same comment as in the oracle diff. `(uint256, bool)` seems like a more natural ordering.

contracts/UFragmentsPolicy.sol
```diff
@@ -81,14 +82,23 @@ contract UFragmentsPolicy is Ownable {
         lastRebaseTimestampSec = now;
         epoch = epoch.add(1);
 
+        bool oracleDataValidity = false;
```
brandoniles: How about `cpiValid`?

contracts/mocks/MockOracle.sol
```diff
@@ -0,0 +1,39 @@
+pragma solidity 0.4.24;
```
brandoniles: If possible, can you make this a move so we don't lose git history?

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

contracts/UFragments.sol
nithinkrishna: 👍

contracts/mocks/MockOracle.sol
nithinkrishna: Done 👍

contracts/UFragmentsPolicy.sol
```diff
@@ -57,12 +55,15 @@ contract UFragmentsPolicy is Ownable {
...
-    uint256 private constant RATE_DECIMALS = 18;
+    // TODO: Explain this parameter
+    uint256 public targetRate;
 
-    uint256 private constant TARGET_RATE = 1 * 10**RATE_DECIMALS;
+    uint256 private constant CPI_DECIMALS = 6;
 
-    int256 private constant TARGET_RATE_SIGNED = int256(TARGET_RATE);
+    uint256 private constant TARGET_CPI = 100 * 10**CPI_DECIMALS;
+    uint256 private constant INITIAL_CPI = 251 * 10**CPI_DECIMALS;
```
nithinkrishna: `INITIAL_CPI` is the CPI at the time of launch. It's used to calculate the `targetRate` at the time of launch. 

```
targetRate = INITIAL_CPI / TARGET_CPI
```

Another way to do this to hard code the `INITIAL_TARGET_RATE`.

contracts/UFragmentsPolicy.sol
```diff
@@ -98,43 +113,42 @@ contract UFragmentsPolicy is Ownable {
...
+     *         modifications are made. RATE_DECIMALS fixed point number.
+     * @param deviationThresholdPerc_ The new exchange rate threshold percentage.
      */
-    function setMinimumVolume(uint256 minimumVolume_)
+    function setDeviationThresholdPerc(uint256 deviationThresholdPerc_)
         external
         onlyOwner
     {
-        minimumVolume = minimumVolume_;
+        deviationThresholdPerc = deviationThresholdPerc_;
```
nithinkrishna: Something to think about.

`deviationThreshold` is cacluated as follows.

```
deviationThreshold = targetRate.mul(deviationThresholdPerc).div(100)
	.div(10 ** RATE_DECIMALS);
```

Should we add a require statement here to ensure that `deviationThresholdPerc` doesn't cause the multiplication `targetRate * deviationThresholdPerc` to overflow?

contracts/mocks/MockOracle.sol
ahnaguib: Most reliable way is to the rename in a separate branch then here, then rebase this one.

contracts/UFragmentsPolicy.sol
```diff
@@ -177,8 +191,11 @@ contract UFragmentsPolicy is Ownable {
     {
         Ownable.initialize(owner);
 
-        deviationThreshold = (5 * TARGET_RATE) / 100;  // 5% of target
-        minimumVolume = 1;
+        targetRate = INITIAL_CPI.mul(10 ** RATE_DECIMALS).div(TARGET_CPI);
+
+        // 5.0%, RATE_DECIMALS Fixed point
+        deviationThresholdPerc = 5 * 10 ** (RATE_DECIMALS);
```
ahnaguib: I think it is cleaner to store it as a fraction not a percentage, so it become 0.05 and removing div(100) at use time

contracts/UFragmentsPolicy.sol
```diff
@@ -81,14 +86,24 @@ contract UFragmentsPolicy is Ownable {
...
+        require(cpiValid);
+
+        targetRate = cpi.mul(10 ** RATE_DECIMALS).div(TARGET_CPI);
+
+        bool rateValid;
         uint256 exchangeRate;
-        uint256 volume;
-        (exchangeRate, volume) = marketOracle.getPriceAnd24HourVolume();
+        (exchangeRate, rateValid) = marketOracle.getData();
+        require(rateValid);
```
ahnaguib: Note: This will change the current behavior when no fresh data exists. Current behavior rebase transaction doesn't fail and supplyDelta is set to 0. With this change rebase transaction will fail.

contracts/UFragmentsPolicy.sol
```diff
@@ -81,14 +86,24 @@ contract UFragmentsPolicy is Ownable {
         lastRebaseTimestampSec = now;
         epoch = epoch.add(1);
 
+        bool cpiValid;
+        uint256 cpi;
+        (cpi, cpiValid) = cpiOracle.getData();
+        require(cpiValid);
+
+        targetRate = cpi.mul(10 ** RATE_DECIMALS).div(TARGET_CPI);
```
nithinkrishna: This can potentially overflow.

contracts/UFragmentsPolicy.sol
```diff
@@ -177,8 +191,11 @@ contract UFragmentsPolicy is Ownable {
     {
         Ownable.initialize(owner);
 
-        deviationThreshold = (5 * TARGET_RATE) / 100;  // 5% of target
-        minimumVolume = 1;
+        targetRate = INITIAL_CPI.mul(10 ** RATE_DECIMALS).div(TARGET_CPI);
```
ahnaguib: storing INITIAL_CPI and computing targetRate in initialize isn't needed.

contracts/UFragmentsPolicy.sol
```diff
@@ -57,18 +55,25 @@ contract UFragmentsPolicy is Ownable {
     // The number of rebase cycles since inception
     uint256 public epoch;
 
-    uint256 private constant RATE_DECIMALS = 18;
-
-    uint256 private constant TARGET_RATE = 1 * 10**RATE_DECIMALS;
+    // TODO: Explain this parameter
+    uint256 public targetRate;
```
ahnaguib: Do we need to have targetRate in storage? Sending it as a parameter to computeSupplyDelta is sufficient.

contracts/UFragmentsPolicy.sol
ahnaguib: s/RATE_DECIMALS/CPI_DECIMALS

contracts/UFragmentsPolicy.sol
```diff
@@ -57,18 +55,25 @@ contract UFragmentsPolicy is Ownable {
...
-    int256 private constant TARGET_RATE_SIGNED = int256(TARGET_RATE);
+    uint256 private constant RATE_DECIMALS = 18;
 
     // Due to the expression in computeSupplyDelta(), MAX_RATE * MAX_SUPPLY must fit into an int256.
     // Both are 18 decimals fixed point numbers.
     uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
+    uint256 private constant CPI_DECIMALS = 6;
```
ahnaguib: I am leaning to have one variable DECIMALS=18 instead of RATE_DECIMALS/CPI_DECIMALS. Interested in hearing the counter argument.

contracts/UFragmentsPolicy.sol
nithinkrishna: Yes, however, the question is do we want to add a require here to prevent rebase from failing due to the overflow?

contracts/UFragmentsPolicy.sol
```diff
@@ -30,19 +30,17 @@ contract UFragmentsPolicy is Ownable {
...
+    IOracle public marketOracle;
 
-    // If the current exchange rate is within this absolute distance from the target, no supply
+    // If the current exchange rate is within this percentage distance from the target, no supply
     // update is performed. Fixed point number--same format as the rate.
-    uint256 public deviationThreshold;
-
-    // The 24hr market volume must be at least this value before any supply adjustments occur.
-    uint256 public minimumVolume;
+    uint256 public deviationThresholdPerc;
```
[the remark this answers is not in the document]
nithinkrishna: Renaming this back to `deviationThreshold` 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: I agree. 👍 
If off-chain services need this value, they can compute it using the data from the CPI oracle and the initial cpi or rate.

contracts/UFragmentsPolicy.sol
nithinkrishna: Makes sense, Having one `DECIMALS` variable makes the code cleaner. 

The only counter-argument is that this data from different oracles won't necessarily be in the same fixed-point format.

contracts/UFragmentsPolicy.sol
```diff
@@ -57,18 +55,25 @@ contract UFragmentsPolicy is Ownable {
...
     // Due to the expression in computeSupplyDelta(), MAX_RATE * MAX_SUPPLY must fit into an int256.
     // Both are 18 decimals fixed point numbers.
     uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
+    uint256 private constant CPI_DECIMALS = 6;
+
+    // CPI-U value AUG 1983 100.2, CPI_DECIMALS Fixed point number
+    uint256 private constant TARGET_CPI = 1002 * 10**(CPI_DECIMALS-1);
```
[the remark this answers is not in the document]
nithinkrishna: I like `BASE_CPI`. Changing it to base cpi. 👍 

Initially, I chose `target` because It felt consistent with `targetRate`, but seems incorrect.

contracts/UFragmentsPolicy.sol
```diff
@@ -57,18 +55,25 @@ contract UFragmentsPolicy is Ownable {
...
 
     // Due to the expression in computeSupplyDelta(), MAX_RATE * MAX_SUPPLY must fit into an int256.
     // Both are 18 decimals fixed point numbers.
     uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
+    uint256 private constant CPI_DECIMALS = 6;
+
+    // CPI-U value AUG 1983 100.2, CPI_DECIMALS Fixed point number
```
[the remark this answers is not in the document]
nithinkrishna: We could choose 100. Or we could choose the CPI pre-WW1 1913 (9.8). I felt that it was more of a product question. 100.2 is more of a placeholder value than anything.

I added the decimal to ensure that our math is consistent.

contracts/UFragmentsPolicy.sol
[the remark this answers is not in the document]
nithinkrishna: @ahnaguib Also suggested this in an earlier comment. Agreed 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: Let's discuss this

contracts/UFragmentsPolicy.sol
nithinkrishna: Let's discuss this

contracts/UFragmentsPolicy.sol
```diff
@@ -57,38 +55,51 @@ contract UFragmentsPolicy is Ownable {
...
-    uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
+    uint256 private constant MAX_RATE = 10**6 * 10**DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
+    // CPI-U value AUG 1983 100.2, DECIMALS Fixed point number
+    uint256 private constant BASE_CPI = 1002 * 10**(DECIMALS-1);
+
+    // CPI-U value JAN 2018 251.712, DECIMALS Fixed point number
+    uint256 private constant INITIAL_CPI = 251712 * 10**(DECIMALS-3);
```
nithinkrishna: Remove this.

contracts/UFragmentsPolicy.sol
```diff
@@ -98,43 +109,42 @@ contract UFragmentsPolicy is Ownable {
...
+     *         modifications are made. DECIMALS fixed point number.
+     * @param deviationThreshold_ The new exchange rate threshold percentage.
      */
-    function setMinimumVolume(uint256 minimumVolume_)
+    function setDeviationThreshold(uint256 deviationThreshold_)
         external
         onlyOwner
     {
-        minimumVolume = minimumVolume_;
+        deviationThreshold = deviationThreshold_;
```
nithinkrishna: Add require here

contracts/UFragmentsPolicy.sol
nithinkrishna: We can't check this if the `targetRate` isn't stored.  Closing this 👍

contracts/UFragmentsPolicy.sol
```diff
@@ -57,38 +56,48 @@ contract UFragmentsPolicy is Ownable {
...
 
     // Due to the expression in computeSupplyDelta(), MAX_RATE * MAX_SUPPLY must fit into an int256.
     // Both are 18 decimals fixed point numbers.
-    uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
+    uint256 private constant MAX_RATE = 10**6 * 10**DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
+    // CPI-U value AUG 1983 100.2, DECIMALS Fixed point number
+    uint256 private constant BASE_CPI = 1002 * 10**(DECIMALS-1);
```
brandoniles: Seems like we should explain why this is DECIMALS -1.

contracts/UFragmentsPolicy.sol
```diff
@@ -57,38 +56,48 @@ contract UFragmentsPolicy is Ownable {
...
+     *      Where DeviationFromTargetRate is (MarketOracleRate - targetRate) / targetRate
+     *      and targetRate is CpiOracleRate / BASE_CPI
      */
     function rebase() external {
         // This comparison also ensures there is no reentrancy.
         require(lastRebaseTimestampSec.add(minRebaseTimeIntervalSec) < now);
         lastRebaseTimestampSec = now;
         epoch = epoch.add(1);
 
+        bool cpiValid;
```
brandoniles: Nit: Let's order this like:
```
uint256 cpi;
bool cpiValid;
```
To match the order below.

contracts/UFragmentsPolicy.sol
```diff
@@ -57,38 +56,48 @@ contract UFragmentsPolicy is Ownable {
...
+        uint256 cpi;
+        (cpi, cpiValid) = cpiOracle.getData();
+        require(cpiValid);
+
+        uint256 targetRate = cpi.mul(10 ** DECIMALS).div(BASE_CPI);
+
+        bool rateValid;
```
brandoniles: Same order comment here.

contracts/UFragmentsPolicy.sol
```diff
@@ -177,8 +185,8 @@ contract UFragmentsPolicy is Ownable {
     {
         Ownable.initialize(owner);
 
-        deviationThreshold = (5 * TARGET_RATE) / 100;  // 5% of target
-        minimumVolume = 1;
+        deviationThreshold = 5 * 10 ** (DECIMALS-2); // 0.05
```
brandoniles: Comment on why this is DECIMALS-2?

contracts/UFragmentsPolicy.sol
```diff
@@ -188,46 +196,41 @@ contract UFragmentsPolicy is Ownable {
...
-        if (withinDeviationThreshold(rate) || !enoughVolume(volume)) {
+        if (withinDeviationThreshold(rate, targetRate)) {
             return 0;
         }
 
-        // (totalSupply * (rate - target)) / target
-        return uFrags.totalSupply().toInt256Safe().mul(
-            rate.toInt256Safe().sub(TARGET_RATE_SIGNED)
-        ).div(TARGET_RATE_SIGNED);
+        // supplyDelta = totalSupply * (rate - targetRate)/targetRate
```
brandoniles:
```suggestion
        // supplyDelta = totalSupply * (rate - targetRate) / targetRate
```

contracts/UFragmentsPolicy.sol
nithinkrishna: I've added an example. Is that sufficient?

contracts/UFragmentsPolicy.sol
nithinkrishna: Based on our offline discussion, we are moving to a scheme where the `tx` fails with invalid data. 

We don't want to be locked out from making supply changes for upto `minRebaseTimeIntervalSec` in the case `rebase` is invoked with no fresh data from either oracle.

contracts/UFragmentsPolicy.sol
```diff
@@ -188,46 +198,41 @@ contract UFragmentsPolicy is Ownable {
...
+            .mul(rate.toInt256Safe().sub(targetRateSigned))
+            .div(targetRateSigned);
     }
 
     /**
      * @param rate The current exchange rate, an 18 decimal fixed point number.
-     * @return If the rate is within the deviation threshold from the target rate, returns true.
-     *         Otherwise, returns false.
+     * @param targetRate The target exchange rate, an 18 decimal fixed point number.
+     * @return If the rate is within the absoluteDeviationThreshold from the targetRate,
```
brandoniles: This comment is probably too much detail. I think the previous version is about right.

contracts/UFragmentsPolicy.sol
```diff
@@ -188,46 +198,41 @@ contract UFragmentsPolicy is Ownable {
...
+    function withinDeviationThreshold(uint256 rate, uint256 targetRate)
         private
         view
         returns (bool)
     {
-        return (rate >= TARGET_RATE && rate.sub(TARGET_RATE) < deviationThreshold)
-            || (rate < TARGET_RATE && TARGET_RATE.sub(rate) < deviationThreshold);
-    }
+        uint256 absoluteDeviationThreshold = targetRate.mul(deviationThreshold)
+            .div(10 ** DECIMALS);
```
brandoniles: I'm confused why we need to .div() here.

If rate.sub(targetRate) is in 18 decimals, don't we want the threshold to be as well?

nithinkrishna: `targetRate`(2.51e18) has 18 decimals, `deviationThreshold`(0.05e18) has 18 aswell. The product (0.125e36) will have 36, so we divide by `1e18`.

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: Kind of an issue. It's 99.9 on July and 100.2 in August. 
https://www.usinflationcalculator.com/inflation/consumer-price-index-and-annual-percent-changes-from-1913-to-2008/

nithinkrishna: Assuming it's a linear increase. 99.9 on July 1st and 100.2 on August 1st. We can choose July 10th as when it was 100.0 ..

contracts/UFragmentsPolicy.sol
```diff
@@ -30,20 +30,19 @@ contract UFragmentsPolicy is Ownable {
...
         uint256 indexed epoch,
         uint256 exchangeRate,
-        uint256 volume24hrs,
+        uint256 cpi,
         int256 requestedSupplyAdjustment
     );
 
     UFragments public uFrags;
-    IMarketOracle public marketOracle;
+    IOracle public cpiOracle;
```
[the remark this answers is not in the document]
nithinkrishna: 👍

contracts/UFragmentsPolicy.sol
```diff
@@ -30,20 +30,23 @@ contract UFragmentsPolicy is Ownable {
...
     );
 
     UFragments public uFrags;
-    IMarketOracle public marketOracle;
 
-    // If the current exchange rate is within this absolute distance from the target, no supply
+    // Cpi oracle provides the current CPI-U data as a 18 decimal fixed point number
+    IOracle public cpiOracle;
+
+    // Market oracle provides the token:USD exchange rate as a 18 decimal fixed point number
```
brandoniles: There's standard notation around exchange rates we can use to decrease any confusion.
e.g.:
```
In the case of the Japanese yen, it's USD/JPY, or dollar to yen. An exchange rate of 100 would mean that 1 dollar equals 100 yen.
```
from [https://www.investopedia.com/terms/e/exchangerate.asp](https://www.investopedia.com/terms/e/exchangerate.asp)

So that should be `token/USD`? How about including this example:
An oracle value of 1.5e18 it would mean 1 Ample is trading for $1.50.

contracts/UFragmentsPolicy.sol
brandoniles: 👍

test/unit/UFragmentsPolicy.js
```diff
@@ -101,62 +114,57 @@ contract('UFragments:setMarketOracle:accessControl', function (accounts) {
...
-  it('should set deviationThreshold', async function () {
-    (await uFragmentsPolicy.deviationThreshold.call()).should.be.bignumber.eq(threshold);
+  it('should set cpiOracle', async function () {
+    await uFragmentsPolicy.setCpiOracle(deployer);
+    expect(await uFragmentsPolicy.cpiOracle.call()).to.eq(deployer);
   });
 });
 
-contract('UFragments:setDeviationThreshold:accessControl', function (accounts) {
+contract('UFragments:setCpiOracle:accessControl', function (accounts) {
```
brandoniles: Should this be 'UFragmentsPolicy:setCpiOracle:accessControl'?

test/unit/UFragmentsPolicy.js
```diff
@@ -423,9 +487,45 @@ contract('UFragmentsPolicy:Rebase', async function (accounts) {
 contract('UFragmentsPolicy:Rebase', async function (accounts) {
   before('setup UFragmentsPolicy contract', setupContracts);
 
-  describe('rate=1', function () {
+  describe('with inflation', function () {
```
brandoniles: This test case name is really confusing, since inflation could refer to USD being devalued, CPI increasing, or Ample supply increasing.

How about 'when cpi increases'?

Same with the 'deflation' case below.

contracts/UFragmentsPolicy.sol
```diff
@@ -30,20 +30,23 @@ contract UFragmentsPolicy is Ownable {
...
-        uint256 volume24hrs,
+        uint256 cpi,
         int256 requestedSupplyAdjustment
     );
 
     UFragments public uFrags;
-    IMarketOracle public marketOracle;
 
-    // If the current exchange rate is within this absolute distance from the target, no supply
+    // Cpi oracle provides the current CPI-U data as a 18 decimal fixed point number
```
brandoniles: CPI-U, seasonally adjusted?

test/unit/UFragmentsPolicy.js
nithinkrishna: 👍

test/unit/UFragmentsPolicy.js
nithinkrishna: I've added one test case and removed another. It's recognized as 2 modifications and not 1 addition and deletion. The file diff in totality should make sense.

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: Yep. `token/USD` makes sense. Added the example aswell.

contracts/UFragmentsPolicy.sol
```diff
@@ -57,38 +62,48 @@ contract UFragmentsPolicy is Ownable {
...
 
     // Due to the expression in computeSupplyDelta(), MAX_RATE * MAX_SUPPLY must fit into an int256.
     // Both are 18 decimals fixed point numbers.
-    uint256 private constant MAX_RATE = 10**6 * 10**RATE_DECIMALS;
+    uint256 private constant MAX_RATE = 10**6 * 10**DECIMALS;
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
+    // CPI-U value July 10th 1983 100, DECIMALS Fixed point number
+    uint256 private constant BASE_CPI = 100 * 10**(DECIMALS);
```
ahnaguib:
```suggestion
    uint256 private constant BASE_CPI = 100 * (10**DECIMALS);
```

PR #112, opened 2019-04-09
Build fix

nithinkrishna:
- Updated frg-runners

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2143,8 +2143,8 @@
       "dev": true
     },
     "frg-ethereum-runners": {
-      "version": "git+https://github.com/ampleforth/frg-ethereum-runners.git#090d0589359a1466c7ad86c64a6a04bc3ef9ac18",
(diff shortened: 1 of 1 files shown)


nithinkrishna:
Waiting on https://github.com/ampleforth/frg-ethereum-runners/pull/4

package-lock.json
```diff
@@ -2143,8 +2143,8 @@
       "dev": true
     },
     "frg-ethereum-runners": {
-      "version": "git+https://github.com/ampleforth/frg-ethereum-runners.git#090d0589359a1466c7ad86c64a6a04bc3ef9ac18",
-      "from": "git+https://github.com/ampleforth/frg-ethereum-runners.git#v1.0.0-alpha",
```
ahnaguib: Should we just use master until we launch? and then assign version numbers?

ahnaguib: Just a suggestion feel free to ignore :)

nithinkrishna: will do that if we make any more changes. Hopefully we wont

PR #115, opened 2019-04-18
add rebase timestamp to logs

truncs:
This is going to save all the block calls since we would be able to use the timestamp from here instead of calling all the blocks individually and getting the timestamp from the block.

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -31,8 +31,8 @@ contract UFragmentsPolicy is Ownable {
         uint256 indexed epoch,
         uint256 exchangeRate,
         uint256 cpi,
-        int256 requestedSupplyAdjustment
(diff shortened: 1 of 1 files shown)


ahnaguib:
To follow our audit process we should submit this to the dev branch instead.

contracts/UFragmentsPolicy.sol
```diff
@@ -31,7 +31,8 @@ contract UFragmentsPolicy is Ownable {
         uint256 indexed epoch,
         uint256 exchangeRate,
         uint256 volume24hrs,
-        int256 requestedSupplyAdjustment
+        int256 requestedSupplyAdjustment,
+        uint256 lastRebaseTimestampSec
```
nithinkrishna: Maybe just call it `timestamp`

truncs: Done!

PR #119, opened 2019-04-25
Initializing base cpi

nithinkrishna:
Base cpi is set during initialization

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -73,15 +73,15 @@ contract UFragmentsPolicy is Ownable {
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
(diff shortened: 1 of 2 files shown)


contracts/UFragmentsPolicy.sol
```diff
@@ -73,15 +73,15 @@ contract UFragmentsPolicy is Ownable {
     // MAX_SUPPLY = MAX_INT256 / MAX_RATE
     uint256 private constant MAX_SUPPLY = ~(uint256(1) << 255) / MAX_RATE;
 
-    // CPI-U value July 10th 1983 100, DECIMALS Fixed point number
-    uint256 private constant BASE_CPI = 100 * (10**DECIMALS);
+    // CPI value at the time of launch, DECIMALS Fixed point number
```
ahnaguib: It's technically not the value at launch but the average of the last released 3 months of PCE price index. not sure how much details of that should we add here.

nithinkrishna: Yea. Let's do another pass for comment fixes.

PR #120, opened 2019-04-26
updated variable ordering

nithinkrishna:
Fixed storage ordering issue with newly added  `baseCpi` variable


contracts/UFragmentsPolicy.sol
```diff
@@ -65,6 +65,9 @@ contract UFragmentsPolicy is Ownable {
     // The number of rebase cycles since inception
     uint256 public epoch;
 
+    // CPI value at the time of launch, DECIMALS Fixed point number
+    uint256 private baseCpi;
```
brandoniles: Conceptually, I'd put this next to deviationThreshold in the code.

nithinkrishna: Done 👍

PR #121, opened 2019-04-29
Comments

brandoniles:
Just making CPI comments less dependent on data source.

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -37,11 +37,10 @@ contract UFragmentsPolicy is Ownable {
 
     UFragments public uFrags;
 
-    // Cpi oracle provides the current CPI-U, seasonally adjusted data,
-    // as a 18 decimal fixed point number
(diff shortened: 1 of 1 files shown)


contracts/UFragmentsPolicy.sol
```diff
@@ -37,11 +37,10 @@ contract UFragmentsPolicy is Ownable {
 
     UFragments public uFrags;
 
-    // Cpi oracle provides the current CPI-U, seasonally adjusted data,
-    // as a 18 decimal fixed point number
+    // Provides the current CPI, as an 18 decimal fixed point number.
```
ahnaguib: for line 47: Should we mention that it's the average of the last released 3 months of PCE price index?

ahnaguib: at line 47: for consistency let's use 18 decimal fixed point number. instead of DECIMALS fixed...

brandoniles: Good catch, done.

PR #127, opened 2019-05-14
Fixed `owner` variable shadowing issue

nithinkrishna:


contracts/UFragments.sol
```diff
@@ -227,16 +227,16 @@ contract UFragments is ERC20Detailed, Ownable {
 
     /**
      * @dev Function to check the amount of tokens that an owner has allowed to a spender.
-     * @param owner The address which owns the funds.
+     * @param owner_ The address which owns the funds.
      * @param spender The address which will spend the funds.
      * @return The number of tokens still available for the spender.
      */
-    function allowance(address owner, address spender)
+    function allowance(address owner_, address spender)
```
ahnaguib: Should we change owner to sender for clarity?

nithinkrishna: Not sure if `sender` would make sense. 

The parameter refers to the owner of the token. So potentially `wallet-x` could check how much allowance `wallet-y` sets for `contract-x` to spend on x's behalf.

https://theethereum.wiki/w/index.php/ERC20_Token_Standard

@brandoniles ?

ahnaguib: sender meaning the token sender not the transaction sender. We can rename to tokenOwner as the wiki example. Owner is mildly confusing with contract owner.

nithinkrishna: 👍

brandoniles: Sender would be confusing I think.

Owner conforms to the convention in the ERC20 standard:
https://theethereum.wiki/w/index.php/ERC20_Token_Standard

Zeppelin, the interface we implement, also uses owner in this context and in the ownable context with no issues.

nithinkrishna: Okay then. Sticking to `owner` for now, we can revisit this later if necessary.

brandoniles:
Not sure I understand the shadowing here... don't we already have a linter that will surface these if they exist?

Ownable has `_owner` has a property.
uFragments has `owner` as a parameter to `initialize()` and `allowance()`

contracts/UFragments.sol
ahnaguib: @brandoniles I recommend tokenOwner over owner as standard here https://theethereum.wiki/w/index.php/ERC20_Token_Standard

PR #129, opened 2019-05-17
TOB recommendations

nithinkrishna:
TOB feedback:

```
1) I think it would be a good idea to add an event in `UFragments.setMonetaryPolicy` so an event is emitted when the `monetaryPolicy` address is updated, so as to be consistent with other event triggers in the contracts.

2) There's also some variable shadowing happening in a few functions. Consider prefixing the function parameter names with an underscore in these cases for clarity. The relevant functions are `UFragmentsPolicy.initialize`, `UFragments.initialize` and `UFragments.allowance`. In each case it's the `owner` parameter shadowing so I'd suggest changing it to `_owner`.

3) I noticed there's a PR ready to be merged in `market-oracle` to enable compiler optimizations (and a similar one was merged earlier this month in `uFragments`). In light of the compiler audit released last fall as well as some recent high-severity bugs related to optimization issues, we currently caution against compiling with optimizations enabled unless the risk of potential latent optimizer bugs is carefully weighed against the gas savings optimizations would provide.
```

### DIFF
--- a/contracts/UFragments.sol
+++ b/contracts/UFragments.sol
@@ -41,6 +41,7 @@ contract UFragments is ERC20Detailed, Ownable {
     event LogRebase(uint256 indexed epoch, uint256 totalSupply);
     event LogRebasePaused(bool paused);
     event LogTokenPaused(bool paused);
+    event LogMonetaryPolicyUpdated(address monetaryPolicy);
 
     // Used for authentication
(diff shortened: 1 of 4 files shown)


ahnaguib:
Separate PRs with descriptive titles is better. Any specific reason we are
bundling them into one?

PR #130, opened 2019-06-03
If oracle data is invalid apply 0 supply change instead of reverting transaction.

ahnaguib:
This was the previous behavior before https://github.com/ampleforth/uFragments/pull/109

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -91,28 +91,29 @@ contract UFragmentsPolicy is Ownable {
         uint256 cpi;
         bool cpiValid;
         (cpi, cpiValid) = cpiOracle.getData();
-        require(cpiValid);
 
         uint256 targetRate = cpi.mul(10 ** DECIMALS).div(baseCpi);
 
         uint256 exchangeRate;
         bool rateValid;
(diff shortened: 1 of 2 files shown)


contracts/UFragmentsPolicy.sol
```diff
@@ -91,28 +91,29 @@ contract UFragmentsPolicy is Ownable {
...
         bool rateValid;
         (exchangeRate, rateValid) = marketOracle.getData();
-        require(rateValid);
 
         if (exchangeRate > MAX_RATE) {
             exchangeRate = MAX_RATE;
         }
 
-        int256 supplyDelta = computeSupplyDelta(exchangeRate, targetRate);
+        int256 supplyDelta = 0;
```
brandoniles: Since we include `exchangeRate` and `cpi` in the rebase log message, we should probably initialize them to 0 the same way we do for `supplyDelta` here.

Functionally it should be the same, but nice to not rely on default values.

ahnaguib: Per our offline discussion, If value is invalid the oracles return zero. It is useful to have the valid value in the log, if only one is invalid.

nithinkrishna:
👍 PR Looks good

PR #133, opened 2019-06-12
Test fix

nithinkrishna:
- Dropped running geth tests on the master build. New change requires evm time manipulation not supported on geth.

- Fixed time-dependent flaky test


test/unit/UFragmentsPolicy.js
```diff
@@ -436,7 +436,7 @@ contract('UFragmentsPolicy:Rebase', async function (accounts) {
       await uFragmentsPolicy.setRebaseTimingParameters(60, 0, 60);
       await chain.waitForSomeTime(60);
       await uFragmentsPolicy.rebase();
-      await chain.waitForSomeTime(72);
```
ahnaguib: Does this value matter at all? (60, 0, 60) means rebase is always possible

PR #144, opened 2020-03-13
AMPL Pool / Balance Registry

brandoniles:
Proof of concept for a simple holder contract that can hold AMPL on behalf of users.

The idea for this is that it will make integrations with AMPL much easier. It's essentially a wrapper token, where users can deposit or withdraw AMPL tokens, and those balances still follow the rebase movements.

Seems like a nice integration piece without the "fixed supply" token we'd throw around earlier.

Still totally untested / uncompiled, but the main functionality should all be there.

We could consider adding deposit(from, to) or withdraw(from, to) style functions as well instead of just the msg.sender versions.

### DIFF
--- a/contracts/AMPLBalanceRegistry.sol
+++ b/contracts/AMPLBalanceRegistry.sol
@@ -1,0 +1,244 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-contracts/contracts/math/SafeMath.sol";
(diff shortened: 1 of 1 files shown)


nithinkrishna:
Can we close this? 

I guess the plan is to roll out an ElasticTokenStandard which the AMPL ERC-20 will confirm to, rather than a wrapper.

brandoniles:
Yeah, don't see a need for this.

PR #146, opened 2020-03-31
Testing rebase invocation from constructor 

nithinkrishna:
Added an additional test to ensure invoking rebase from within a contract constructor fails.

### DIFF
--- a/contracts/mocks/ConstructorRebaseCallerContract.sol
+++ b/contracts/mocks/ConstructorRebaseCallerContract.sol
@@ -1,0 +1,16 @@
+pragma solidity 0.4.24;
+
+import "../UFragmentsPolicy.sol";
+
+
+contract ConstructorRebaseCallerContract {
+    UFragmentsPolicy private _policy;
+
+    constructor(address policy) public {
+        _policy = UFragmentsPolicy(policy);
+        // Take out a flash loan.
+        // Do something funky...
+        _policy.rebase();  // should fail
+        // pay back flash loan.
+    }
+}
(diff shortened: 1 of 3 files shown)


contracts/mocks/ConstructorRebaseCallerContract.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity 0.4.24;
+
+import "../UFragmentsPolicy.sol";
+
+
+contract ConstructorRebaseCallerContract {
+    UFragmentsPolicy private _policy;
```
brandoniles: Don't need this property since it's used in the function it's passed into.

test/unit/UFragmentsPolicy.js
```diff
@@ -292,6 +293,14 @@ contract('UFragmentsPolicy:Rebase:accessControl', async function (accounts) {
       ).to.be.true;
     });
   });
+
+  describe('when rebase called by a contract which is being constructed', function () {
+    it('should fail', async function () {
+      expect(
+        await chain.isEthException(ConstructorRebaseCallerContract.new(uFragmentsPolicy.address))
```
brandoniles: We should use the same pattern as above, but this one's shorter. Might as well update the other test function to not store `rebaseCallerContract` etc.

contracts/mocks/ConstructorRebaseCallerContract.sol
nithinkrishna: Done 👍

test/unit/UFragmentsPolicy.js
nithinkrishna: Done 👍

PR #151, opened 2020-05-15
Certik Audit Recommendations

brandoniles:
Exhibit 1-
Remove Orchestrator address initialization in the policy.

Exhibit 2-
Reorder Transaction struct elements to pack values and save storage.

Exhibit 3-
No action. I'm ok with leaving the index in the `TransactionFailed` event, since risk is minimal and it aids potential debugging.

Exhibit 4-
No action needed.

Exhibit 5-
Removed unnecessary `delete` call.

Exhibit 6-
Removed `transferValueWei` parameter in external_call.

Exhibit 7-
Remove `dataLength` parameter in external_call. Load directly from array's storage.

Exhibit 8-
No action. It doesn't save gas, and I think it's more legible as is.

Exhibit 9-
Renamed `transactionsLength` to `transactionsSize` to follow market-oracle repo's convention.

Exhibit 10-
No action needed, since we mitigated Exhibit 2.

### DIFF
--- a/contracts/Orchestrator.sol
+++ b/contracts/Orchestrator.sol
@@ -52,7 +52,7 @@ contract Orchestrator is Ownable {
             Transaction storage t = transactions[i];
             if (t.enabled) {
                 bool result =
-                    externalCall(t.destination, 0, t.data.length, t.data);
+                    externalCall(t.destination, t.data.length, t.data);
                 if (!result) {
                     emit TransactionFailed(t.destination, i, t.data);
                 }
@@ -119,12 +119,11 @@ contract Orchestrator is Ownable {
     /**
      * @dev wrapper to call the encoded transactions on downstream consumers.
      * @param destination Address of destination contract.
-     * @param transferValueWei ETH value to send, in wei.
      * @param dataLength Size of data param.
      * @param data The encoded data payload.
      * @return True on success
      */
-    function externalCall(address destination, uint transferValueWei, uint dataLength, bytes data)
+    function externalCall(address destination, uint dataLength, bytes data)
         internal
         returns (bool)
     {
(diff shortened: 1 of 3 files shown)


nithinkrishna:
Looks good. Thanks 👍 

Agree you call wrt Exhibit 3 and 8.

Have a question regarding Exhibit 10. 

Wouldn't renaming the `enabled` variable to `disabled` be better? 

Seems like it knocks off one write operation during transaction creation.
Since disabling is much rarer than creation overall would reduce gas consumption.

```
In Solidity, if a value remains uninitialized it will cost less gas-cost to retain rather than an initialized value. As such, it is possible to optimize this step by renaming the variable from “enabled” to “disabled” and setting the default value to “false”. This will cause transactions to cost less gas to create but increase the gas cost of ~~creating~~ updating them. As disabling transactions is anticipated to be seldomly used in contrast to creating them, applying the aforementioned change would lead to an overall reduction in gas consumption.
```
(There might have been a typo in their report)

nithinkrishna:
LGTM 👍

contracts/Orchestrator.sol
```diff
@@ -90,7 +90,6 @@ contract Orchestrator is Ownable {
             transactions[index] = transactions[transactions.length - 1];
         }
 
-        delete transactions[transactions.length - 1];
```
ahnaguib, line 93: That's hidden and good to know! we have this delete pattern in multiple places in our codebase.

contracts/Orchestrator.sol
```diff
@@ -147,9 +144,9 @@ contract Orchestrator is Ownable {
 
 
                 destination,
-                transferValueWei,
+                0, // transfer value in wei
```
ahnaguib, line 147: I thought we already did this when I left a comment for it in the original PR.

PR #159, opened 2020-07-28
Remove pause functions

brandoniles:
https://github.com/ampleforth/RFCs/blob/master/RFCs/rfc-3.md


contracts/UFragments.sol
```diff
@@ -51,19 +49,8 @@ contract UFragments is ERC20Detailed, Ownable {
...
-    modifier whenRebaseNotPaused() {
-        require(!rebasePaused);
-        _;
-    }
-
-    modifier whenTokenNotPaused() {
-        require(!tokenPaused);
-        _;
-    }
+    bool public rebasePausedDeprecated;
```
nithinkrishna, line 52: As far I recall making this private should not affect anything. It will also remove these variable from the ABI.

But If we want to be extra-cautious, I'm okay with renaming.  👍 

Leaving it to you to decide

brandoniles: From the docs, only order and type determine storage layout: https://solidity.readthedocs.io/en/v0.4.24/miscellaneous.html#layout-of-state-variables-in-storage

Also for most recent v0.7.0: https://solidity.readthedocs.io/en/v0.7.0/internals/layout_in_storage.html?highlight=storage%20layout

This is corroborated by OpenZeppelin's high level guide: https://docs.openzeppelin.com/upgrades/2.6/writing-upgradeable

So I'll move to `private`, good idea.

thegostep: If you want additional certainty on stability of storage layout it could be a good idea to run integration tests on forked mainnet state

contracts/UFragments.sol
```diff
@@ -175,8 +137,8 @@ contract UFragments is ERC20Detailed, Ownable {
         ERC20Detailed.initialize("Ampleforth", "AMPL", uint8(DECIMALS));
         Ownable.initialize(owner_);
 
-        rebasePaused = false;
-        tokenPaused = false;
+        rebasePausedDeprecated = false;
+        tokenPausedDeprecated = false;
```
thegostep, line 141: the initialize function should be removed entirely when deploying a new template (since it is never called)

contracts/UFragments.sol
brandoniles: Good ideas thanks!

We have a script that dumps storage state so we can compare before and after.

We also have a mainnet-staging that we test with before mainnet production.

contracts/UFragments.sol
brandoniles: Good to know.

There could be a future where we do a separate non-upgrade deployment. Could be on Ethereum or an alternate chain that works with solidity tools. In that case, it'd be useful to have it there already.

thegostep: I see - in that case I would recommend keeping as a comment since you want to avoid deploying unreachable code

PR #160, opened 2020-08-04
Update README.md

ggq89:


nithinkrishna:
Hey @ggq89 Thanks for your contribution. 

Going forward I think there will be multiple research papers and `https://www.ampleforth.org/paper/` will point to a landing page with all of them ..

PR #161, opened 2020-08-11
add peripheral reverts to orchestrator

thegostep:
Policy for tolerated peripheral:
- no upgrade (because then calling arbitrary code)
- no reverts with no message (because reverts the batch)


contracts/Orchestrator.sol
```diff
@@ -46,35 +48,107 @@ contract Orchestrator is Ownable {
...
+
+    /**
+     * @notice Get the revert message and code from a call.
+     * @param res bytes Response of the call.
+     * @return revertMessage string Revert message.
+     * @return revertCode bytes32 Revert code.
+     */
+	function _getRevertMsg(bytes memory res) internal pure returns (string memory revertMessage, bytes32 revertCode) {
+        // If there is no prefix to the revert reason, we know it was an OOG error
+        if (res.length == 0) {
```
thegostep, line 110: check if assert or any other scenario can trigger this

contracts/Orchestrator.sol
```diff
@@ -46,35 +48,107 @@ contract Orchestrator is Ownable {
...
+            revertMessage = "Transaction out of gas";
+        }
+		// If the revert reason length is less than 68, then the transaction failed silently (without a revert message)
+		else if (res.length < 68) {
```
thegostep, line 114: verify that reverts before v0.4 triggers this

contracts/Orchestrator.sol
```diff
@@ -46,35 +48,107 @@ contract Orchestrator is Ownable {
...
+
     /**
      * @notice Adds a transaction that gets called for a downstream receiver of rebases
      * @param destination Address of contract destination
      * @param data Transaction data payload
+     * @param gasLimit Transaction gas limit
+     * @param revertOKs Transaction approved revert codes
      */
-    function addTransaction(address destination, bytes data)
+    function addTransaction(address destination, bytes data, uint256 gasLimit, bytes32[] revertOKs)
```
thegostep, line 135: define policy of how to audit peripheral
- no upgrade
- no gas revert spoofing

contracts/Orchestrator.sol
```diff
@@ -46,35 +48,107 @@ contract Orchestrator is Ownable {
...
+
+        // validate sufficient gas left
+        require(gasleft() > transaction.gasLimit);
+
+        // perform external call
+        // todo: @thegostep solc v0.4 does not return revert string on call
+        // todo: @thegostep solc v0.4 does not support specifying gaslimit on call
+        // decide if upgrade solc or implement in assembly
+        // https://solidity.readthedocs.io/en/v0.4.24/units-and-global-variables.html#address-related
+        (bool success, bytes memory res) = address(transaction.destination).call(transaction.data);
```
thegostep, line 78: check if can upgrade solc version to latest ToB version without breaking everything
if requires re-writing test suite, consider assembly

contracts/Orchestrator.sol
thegostep: - a revert with no message would trigger the out of gas error check

thegostep: - avoid external call into solc <0.4

contracts/Orchestrator.sol
thegostep: it does not

contracts/Orchestrator.sol
thegostep: yes it does

thegostep:
It is not clear to me how this change actually helps mitigate any potential DOS. Since we are forwarding all remaining gas to the external call, an OOG error would revert the batch regardless unless maybe if it was the last call in the batch

brandoniles:
> It is not clear to me how this change actually helps mitigate any potential DOS. Since we are forwarding all remaining gas to the external call, an OOG error would revert the batch regardless unless maybe if it was the last call in the batch

This design doesn't keep downstream calls from DoSing rebase by using up all the gas.
But it does prevent callers from selectively truncating downstream calls through gas manipulation.

If we want to mitigate DoS with this design, it must be done through manually auditing the code before it's added to the transactions list. We'd require:
- No upgradability
- Limit on gas usage, i.e. no unbounded operations

If we wanted a more generalized design, we'd add another parameter to each transaction representing how much gas to forward to it. If a tx uses more than its allotment, we let it fail and continue. Before we execute anything, we enforce the user has provided at least enough gas to cover (1) our rebase + code (2) Sum of all tx max gas amounts.

Even with the generalized approach, we'd still audit code before addition anyway, but it would use more gas overall. Question is whether the extra safety stops are worth it or not. I think right now, we'd err on simplicity.

thegostep:
@nithinkrishna bumping the version of truffle broke a lot of tests - am not sure if it is worthwhile to fix

PR #164, opened 2020-08-28
buidler-refactor

thegostep:
- [x] migrate tests
  - [x] convert to typescript
  - [x] add use of fixtures https://ethereum-waffle.readthedocs.io/en/latest/fixtures.html
  - [x] add use of mocks https://ethereum-waffle.readthedocs.io/en/latest/mock-contract.html
- [x] update deploy scripts
  - [x] remove manual steps
  - [x] add manager-owner pattern
  - [x] add zeppelin sdk-cli compatibility
- [ ] update CI
  - [x] update travis
  - [x] update precommit hooks: https://prettier.io/docs/en/precommit.html
  - [ ] add crytic: https://github.com/crytic/slither/wiki/Upgradeability-Checks


nithinkrishna:
@thegostep can you make the test file renames register as moves rather than deletion/additions.


https://stackoverflow.com/questions/433111/how-to-make-git-mark-a-deleted-and-a-new-file-as-a-file-move

thegostep:
> @thegostep can you make the test file renames register as moves rather than deletion/additions.

I took a crack at this - unfortunately I don't think it is possible for git to track diffs between js and ts files

nithinkrishna (approved):

nithinkrishna:
Looks good. But I'd say keep it in a sperate branch (development) for now.

thegostep:
Here is a great tutorial on process for upgrades: https://forum.openzeppelin.com/t/openzeppelin-upgrades-step-by-step-tutorial-for-hardhat/3580

.gitignore
```diff
@@ -72,3 +72,10 @@ scDebugLog
 coverage.json
 coverage/
 coverageEnv/
+
+node_modules
```
brandoniles, line 76: I also see that package-lock.json is removed. Will we still get reproducible builds?

brandoniles (commented):
Thanks Stephane, I will always support typed language!

I see that package-lock.json is removed and node_modules is added to gitignore. Will we still get reproducible builds?

AFAICT, the only changes to the solidity files are formatting. I assume this matches the new linter settings. As a general note, I'd recommend separating the code changes from the test changes when the behavior is meant to be the same. It gives more confidence that neither introduces problems on their own.

thegostep:
> I see that package-lock.json is removed and node_modules is added to gitignore. Will we still get reproducible builds?

Yes - I switched the package manager from npm to yarn. See yarn.lock for version locking

thegostep:
> AFAICT, the only changes to the solidity files are formatting. I assume this matches the new linter settings. As a general note, I'd recommend separating the code changes from the test changes when the behavior is meant to be the same. It gives more confidence that neither introduces problems on their own.

Fair - I can split into seperate PRs if you prefer. Just say the word ✨

thegostep:
following @brandoniles suggestion, this PR is superseded by #176 and #177

PR #170, opened 2020-11-10
Sigmoid rebase curve

ahnaguib:


contracts/UFragmentsPolicy.sol
```diff
@@ -85,17 +84,19 @@ contract UFragmentsPolicy is Ownable {
...
+    int256 public rebaseFunctionUpperPercentage;
+    int256 public rebaseFunctionGrowth;
+
     modifier onlyOrchestrator() {
         require(msg.sender == orchestrator);
         _;
     }
 
     /**
      * @notice Initiates a new rebase operation, provided the minimum time period has elapsed.
```
nithinkrishna, line 99: Maybe update the comment with the new equation?

contracts/UFragmentsPolicy.sol
```diff
@@ -172,6 +170,22 @@ contract UFragmentsPolicy is Ownable {
...
 
+    function setRebaseFunctionGrowth(int256 rebaseFunctionGrowth_)
+        external
+        onlyOwner
+    {
+        require(rebaseFunctionGrowth_ >= 0);
+        rebaseFunctionGrowth = rebaseFunctionGrowth_;
+    }
+
+    function setRebaseFunctionUpperPercentage(int256 rebaseFunctionUpperPercentage_)
```
brandoniles, line 181: We should add a setter for the lower now, too.

contracts/UFragmentsPolicy.sol
```diff
@@ -239,10 +237,12 @@ contract UFragmentsPolicy is Ownable {
     {
         Ownable.initialize(owner_);
 
-        // deviationThreshold = 0.05e18 = 5e16
-        deviationThreshold = 5 * 10 ** (DECIMALS-2);
+        deviationThreshold = 0;
+
+        rebaseFunctionGrowth = int256(4 * (10**DECIMALS));
+        rebaseFunctionUpperPercentage = int256(11 * (10**(DECIMALS-2))); // .11
+        rebaseFunctionLowerPercentage = int256((-11) * int256(10**(DECIMALS-2))); // .11
```
brandoniles, line 244: "-.11"

contracts/UFragmentsPolicy.sol
```diff
@@ -264,6 +264,40 @@ contract UFragmentsPolicy is Ownable {
         );
     }
 
+    /**
+     * @return Computes the percentage of supply to be added or removed.
```
brandoniles, line 268:
```suggestion
     * @return The percentage of supply to be added or removed.
```
Usually the @return goes last in the NatSpec block.

contracts/UFragmentsPolicy.sol
```diff
@@ -264,6 +264,40 @@ contract UFragmentsPolicy is Ownable {
...
+     * Using the function in https://github.com/ampleforth/AIPs/blob/master/AIPs/aip-5.md
+     * @param normalizedRate a DECIMALS decimal fixed point number.
```
brandoniles, line 270: I think `normalizedRate` needs an explanation. It could be in the @dev block to keep the @param line short.

contracts/UFragmentsPolicy.sol
```diff
@@ -264,6 +264,40 @@ contract UFragmentsPolicy is Ownable {
...
+        pure
+        returns (int256)
+    {
+        int256 delta;
+
+        delta = (normalizedRate.sub(ONE));
+
+        // Compute: (Upper-Lower)/(1-(Upper/Lower)/2^(Growth*delta))) + Lower
+
+        int256 exponent = growth.mul(delta).div(ONE);
```
brandoniles, line 284: @thegostep Is using a fixed point library to handle the `ONES`. You two should sync to see if it makes sense to share it.

contracts/lib/SafeMathInt.sol
```diff
@@ -102,4 +102,45 @@ library SafeMathInt {
         require(a != MIN_INT256);
         return a < 0 ? -a : a;
     }
+
+    /**
+    * @dev Computes 2^exp with limited precision where 0 <= exp <= 100 * one
+    * @param one 1.0 represented in the same fixed point number format as exp
```
brandoniles, line 108: Can you have 2 @param's and 1 @return ?

brandoniles (commented):
Looks good, mostly cosmetic things. I'd sync with Stephane about his fixed point library though.

contracts/UFragmentsPolicy.sol
ahnaguib: Done.

contracts/UFragmentsPolicy.sol
ahnaguib: Good catch!

contracts/UFragmentsPolicy.sol
ahnaguib: Done.

contracts/UFragmentsPolicy.sol
ahnaguib: added rate/target in the param

contracts/UFragmentsPolicy.sol
ahnaguib: @thegostep Which library did you end up using, the one I linked or a different one?

contracts/lib/SafeMathInt.sol
ahnaguib: looks good? @brandoniles

contracts/UFragmentsPolicy.sol
thegostep: My current implementation uses https://github.com/HQ20/contracts/tree/master/contracts/math

I like that the operations behave like SafeMath but I have not settled on it.

Here are viable alternatives:
- https://github.com/abdk-consulting/abdk-libraries-solidity
- https://github.com/albertocuestacanada/DecimalMath/blob/master/contracts/DecimalMath.sol
- https://github.com/compound-finance/compound-protocol/blob/master/contracts/Exponential.sol

PR #177, opened 2020-12-23
Add support for and apply auto code formatting

thegostep:


contracts/UFragments.sol
```diff
@@ -150,23 +143,15 @@ contract UFragments is ERC20Detailed, Ownable {
     /**
      * @return The total number of fragments.
      */
-    function totalSupply()
-        public
-        view
-        returns (uint256)
-    {
+    function totalSupply() public view returns (uint256) {
```
ahnaguib, line 146: I prefer the existing format for function modifiers. Don't consider it a blocker though.

thegostep: I don't think that is configurable unfortunately, I would recommend sticking to what is supported under prettier-solidity: https://github.com/prettier-solidity/prettier-plugin-solidity

brandoniles (approved):

contracts/UFragments.sol
ahnaguib: SGTM

ahnaguib (approved):

Issue #179, opened 2021-01-07
Expose scaled balances similar to aTokens


nithinkrishna:
-> https://docs.aave.com/developers/the-core-protocol/atokens#scaledbalanceof
-> https://docs.aave.com/developers/the-core-protocol/atokens#scaledtotalsupply

PR #180, opened 2021-01-08
Elastic token standard: Interface update

nithinkrishna:
Adding public view 2 methods to the AMPL ERC-20, similar to [AAVE's aToken](https://docs.aave.com/developers/the-core-protocol/atokens) to expose the internal user balance and total supply.

 * `scaledBalanceOf` : internal gon balance
 * `scaledTotalSupply` : total number of gons

Fixes: https://github.com/ampleforth/uFragments/issues/179

### DIFF
--- a/contracts/UFragments.sol
+++ b/contracts/UFragments.sol
@@ -156,6 +156,21 @@ contract UFragments is ERC20Detailed, Ownable {
     }
 
     /**
+     * @param who The address to query.
+     * @return The gon balance of the specified address.
+     */
+    function scaledBalanceOf(address who) public view returns (uint256) {
+        return _gonBalances[who];
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


ahnaguib (approved):

brandoniles (approved):
LGTM

thegostep (approved):
lgtm - I am still partial to exposing gons with an ERC20 interface through a separate address for improved composability, but that admittedly makes things a bit more funky

nithinkrishna:
@thegostep yea we've talked about that. I dont see why we can't do that independent of this

thegostep:
> @thegostep yea we've talked about that. I dont see why we can't do that independent of this

makes sense - though it does require a token upgrade to deploy

PR #181, opened 2021-01-09
Permit [eip-2612] support for AMPL

nithinkrishna:
Gasless approvals with permit. 
https://eips.ethereum.org/EIPS/eip-2612

Reference implementations
* Openzeppelin: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/drafts/ERC20Permit.sol
* AToken: https://github.com/ampleforth/protocol-v2/blob/master/contracts/protocol/tokenization/AToken.sol

There's one major discussion point. 

As per the standard openzeppelin implementation, the domain separator is computed in the [contract's constructor](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/drafts/EIP712.sol#L46). However since we have a live contract instance, we cant do that.

In the current approach, I've hardcoded the values that go into the domain separator creation based on the production config. (Address, chainId and token name).

If we dont want to hardcode this, we could create a function which we invoke after upgrade which sets the domain separator after reading the contract's store (more general solution).


contracts/UFragments.sol
```diff
@@ -255,4 +291,53 @@ contract UFragments is ERC20Detailed, Ownable {
...
+                            spender,
+                            value,
+                            currentValidNonce,
+                            deadline
+                        )
+                    )
+                )
+            );
+
+        require(owner == ecrecover(digest, v, r, s));
```
thegostep, line 336: add revert message?

test/unit/UFragments.ts
```diff
@@ -1,15 +1,53 @@
 import { ethers, upgrades } from 'hardhat'
-import { Contract, Signer, BigNumber } from 'ethers'
+import { Contract, Signer, BigNumber, BigNumberish } from 'ethers'
 import { expect } from 'chai'
+import { fromRpcSig } from 'ethereumjs-util'
+import { TypedDataUtils, signTypedMessage, TypedMessage } from 'eth-sig-util'
```
thegostep, line 5: Could use ethers for typed data signing: https://docs.ethers.io/v5/api/signer/#Signer-signTypedData

contracts/UFragments.sol
```diff
@@ -155,6 +183,14 @@ contract UFragments is ERC20Detailed, Ownable {
         return _gonBalances[who].div(_gonsPerFragment);
     }
 
+    /**
+     * @param who The address to query.
+     * @return The number of successful permits by the specified address.
+     */
+    function nonces(address who) public view returns (uint256) {
```
thegostep, line 190: set to external?

thegostep (approved):
lgtm - though was only able to take a cursory look

contracts/UFragments.sol
nithinkrishna: Dont think revert messages are possible in 0.4.24

contracts/UFragments.sol
nithinkrishna: kept it consistent with `balanceOf` and `totalSupply`. Worthwhile changing all of them to `external`?

thegostep: yeah - it will give a (small) gas saving

test/unit/UFragments.ts
nithinkrishna: Done 👍

contracts/UFragments.sol
nithinkrishna: https://github.com/ampleforth/uFragments/issues/183

contracts/UFragments.sol
```diff
@@ -76,6 +76,34 @@ contract UFragments is ERC20Detailed, Ownable {
     // it's fully paid.
     mapping(address => mapping(address => uint256)) private _allowedFragments;
 
+    // EIP-2612: permit – 712-signed approvals
+    // https://eips.ethereum.org/EIPS/eip-2612
+    bytes public constant EIP712_REVISION = "1";
```
brandoniles, line 81: ` = bytes('1');` to be explicit? This is what AToken does.

nithinkrishna: that doesn't work in solidity 0.4.24 :/

brandoniles (approved):
Thanks!

nithinkrishna:
Will wait for @ahnaguib to review this

contracts/UFragments.sol
```diff
@@ -278,4 +313,39 @@ contract UFragments is ERC20Detailed, Ownable {
...
+        bytes32 r,
+        bytes32 s
+    ) public {
+        require(block.timestamp <= deadline);
+
+        uint256 ownerNonce = _nonces[owner];
+        bytes32 permitDataDigest =
+            keccak256(abi.encode(PERMIT_TYPEHASH, owner, spender, value, ownerNonce, deadline));
+        bytes32 digest =
+            keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR, permitDataDigest));
```
nithinkrishna, line 342: @ahnaguib How OpenZeppelin implements this

https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/drafts/EIP712Upgradeable.sol#L88

https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/drafts/ERC20PermitUpgradeable.sol#L50

package.json
```diff
@@ -39,7 +39,8 @@
     "bignumber.js": "^9.0.0",
     "chai": "^4.2.0",
     "ethereum-waffle": "^3.2.1",
-    "ethers": "^5.0.24",
+    "ethereumjs-util": "^7.0.7",
```
thegostep, line 42: doesn't seem like this import is required anymore

nithinkrishna, line 42:
```suggestion
```

test/utils/signatures.ts
```diff
@@ -0,0 +1,121 @@
+// https://github.com/albertocuestacanada/ERC20Permit/blob/master/utils/signatures.ts
+import {
+  keccak256,
+  defaultAbiCoder,
+  toUtf8Bytes,
+  solidityPack,
+  splitSignature,
+} from 'ethers/lib/utils'
+import { ecsign } from 'ethereumjs-util'
```
nithinkrishna, line 9:
```suggestion
```

brandoniles (changes requested):
Until we address the hardcoded separator issue.

nithinkrishna:
Holding off on this until we bump up the token solidity compiler version

thegostep:
OZ released Permit contract today: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.4.0/contracts/drafts/ERC20Permit.sol

nithinkrishna:
They've had it in drafts for a while now

nithinkrishna:
@brandoniles I've updated to the new implementation which queries the chainId everytime

contracts/UFragments.sol
```diff
@@ -174,6 +189,42 @@ contract UFragments is ERC20Detailed, Ownable {
...
+     * @return The number of successful permits by the specified address.
+     */
+    function nonces(address who) public view returns (uint256) {
+        return _nonces[who];
+    }
+
+    /**
+     * @return The underlying blockchain's chainID.
+     */
+    function getChainId() public view returns (uint256) {
```
brandoniles, line 202: Does this need to be public?

Also, basic question but--does this being its own function use more gas than it would if it were just in the body of `DOMAIN_SEPARATOR()`?

nithinkrishna: Hmm. Not necessarily. But if it is public, off-chain services can look up all the attributes to compute the domain_seperator from the token contract's interface.

```
keccak256(
    abi.encode(
        uFragments.EIP712_DOMAIN(),
        keccak256(bytes(uFragments.name())),
        keccak256(bytes(uFragments.EIP712_REVISION())),
        uFragments.getChainId(),
        uFragments.address
    )
)
```

nithinkrishna: Makes sense.

brandoniles (approved):
LGTM, but wait for at least @thegostep 's approval also.

thegostep (approved):
lgtm

Issue #183, opened 2021-01-11
ERC-20 use external instead of public


nithinkrishna:

nithinkrishna:
https://github.com/ampleforth/uFragments/pull/186

PR #186, opened 2021-01-12
Updated function access modifiers 

### LINKED ISSUE #185: `scaledTotalSupply` can be pure

nithinkrishna:
fixes #185 #183

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -249,7 +249,7 @@ contract UFragmentsPolicy is Ownable {
      * @return Computes the total supply adjustment in response to the exchange rate
      *         and the targetRate.
      */
-    function computeSupplyDelta(uint256 rate, uint256 targetRate) private view returns (int256) {
+    function computeSupplyDelta(uint256 rate, uint256 targetRate) internal view returns (int256) {
         if (withinDeviationThreshold(rate, targetRate)) {
             return 0;
         }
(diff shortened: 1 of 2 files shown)


ahnaguib (approved):

contracts/UFragmentsPolicy.sol
```diff
@@ -249,7 +249,7 @@ contract UFragmentsPolicy is Ownable {
      * @return Computes the total supply adjustment in response to the exchange rate
      *         and the targetRate.
      */
-    function computeSupplyDelta(uint256 rate, uint256 targetRate) private view returns (int256) {
+    function computeSupplyDelta(uint256 rate, uint256 targetRate) internal view returns (int256) {
```
brandoniles, line 252: Confusingly, I think `private` -> `internal` has both a gas change (jump w/ no memory clear) and a functional change (visibility within derived contracts). Just pointing out the change to visibility, but I think that's ok.

https://docs.soliditylang.org/en/v0.8.0/control-structures.html?highlight=internal#internal-function-calls
https://docs.soliditylang.org/en/v0.8.0/contracts.html?highlight=private#visibility-and-getters

brandoniles (approved):

contracts/UFragmentsPolicy.sol
nithinkrishna: Noted. Yea. That's weird

PR #189, opened 2021-01-14
Add getGlobalAmpleforthEpochAndAMPLSupply

### LINKED ISSUE #188: add getter on policy which returns epoch and global supply

nithinkrishna:
Added a getter on the Policy contract which returns the `globalAmpleforthEpoch` and `globalAMPLSupply`. The MonetaryPolicy (on the base chain) and the XCAmpleController on the satellite chains will implement this method.

When rebase is propagated between chains, its these 2 values which are transmitted. The method allows consumers to read both values atomically.  

fixes #188

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -211,6 +211,20 @@ contract UFragmentsPolicy is Ownable {
     }
 
     /**
+     * @notice A multi-chain AMPL interface method. The Ampleforth monetary policy contract
+     *         on the base-chain and XCAmpleController contracts on the satellite-chains
+     *         implement this method. It returns what the current contract believes to be
+     *         the globalAmpleforthEpoch and globalAMPLSupply.
+     * @return globalAmpleforthEpoch The current epoch number.
+     * @return globalAMPLSupply The total supply at the current epoch.
+     */
+    function getGlobalAmpleforthEpochAndAMPLSupply() external view returns (uint256, uint256) {
+        uint256 globalAmpleforthEpoch = epoch;
+        uint256 globalAMPLSupply = uFrags.totalSupply();
+        return (globalAmpleforthEpoch, globalAMPLSupply);
+    }
+
+    /**
      * @dev ZOS upgradable contract initialization method.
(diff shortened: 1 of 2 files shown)


contracts/UFragmentsPolicy.sol
```diff
@@ -210,6 +210,20 @@ contract UFragmentsPolicy is Ownable {
...
+     * @notice A multi-chain AMPL interface method. The Ampleforth monetary policy contract
+     *         on the base-chain and XCAmpleController contracts on the satellite-chains
+     *         implement this method. It returns what the current contract believes to be
+     *         the globalAmpleforthEpoch and globalAMPLSupply.
+     * @return globalAmpleforthEpoch The current epoch number.
+     * @return globalAMPLSupply The total supply at the current epoch.
+     */
+    function getGlobalAmpleforthEpochAndAMPLSupply() external view returns (uint256, uint256) {
+        uint256 globalAmpleforthEpoch = epoch;
+        uint256 globalAMPLSupply = uFrags.totalSupply();
```
brandoniles, line 223: Is this really more efficient than doing...
`return (epoch, uFrags.totalSupply());`

contracts/UFragmentsPolicy.sol
```diff
@@ -210,6 +210,20 @@ contract UFragmentsPolicy is Ownable {
         rebaseWindowLengthSec = rebaseWindowLengthSec_;
     }
 
+    /**
+     * @notice A multi-chain AMPL interface method. The Ampleforth monetary policy contract
+     *         on the base-chain and XCAmpleController contracts on the satellite-chains
+     *         implement this method. It returns what the current contract believes to be
```
brandoniles, line 216: "It returns what the...." 
->
"It atomically returns two values: what the current..."

brandoniles (approved):

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

contracts/UFragmentsPolicy.sol
nithinkrishna: 👍

PR #190, opened 2021-01-19
Public Contract interfaces

nithinkrishna:
* Checked in standard public interface definition for, AMPL token, Ampleforth monetary policy and Orchestrator.

### DIFF
--- a/contracts/interfaces/IAMPL.sol
+++ b/contracts/interfaces/IAMPL.sol
@@ -1,0 +1,63 @@
+pragma solidity 0.4.24;
+
+// Interface definition for the AMPL - ERC20 token on Ethereum (the base-chain)
+interface IAMPL {
+    // ERC20
+    function totalSupply() external view returns (uint256);
+
+    function balanceOf(address who) external view returns (uint256);
+
+    function allowance(address owner_, address spender) external view returns (uint256);
+
+    function transfer(address to, uint256 value) external returns (bool);
+
+    function approve(address spender, uint256 value) external returns (bool);
+
+    function increaseAllowance(address spender, uint256 addedValue) external returns (bool);
+
+    function decreaseAllowance(address spender, uint256 subtractedValue) external returns (bool);
+
+    function transferFrom(
+        address from,
+        address to,
+        uint256 value
+    ) external returns (bool);
+
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    event Approval(address indexed owner, address indexed spender, uint256 value);
+
+    // EIP-2612
+    function permit(
+        address owner,
+        address spender,
+        uint256 value,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external;
+
+    function nonces(address owner) external view returns (uint256);
+
+    function DOMAIN_SEPARATOR() external view returns (bytes32);
+
+    // Elastic token interface
+    function scaledBalanceOf(address who) external view returns (uint256);
+
+    function scaledTotalSupply() external pure returns (uint256);
+
+    function transferAll(address to) external returns (bool);
+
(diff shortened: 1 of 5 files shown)


brandoniles (commented):
Should we include the Orchestrator? That has the main rebase callsite

contracts/interfaces/IAMPL.sol
```diff
@@ -0,0 +1,57 @@
...
+    function scaledBalanceOf(address who) external view returns (uint256);
+
+    function scaledTotalSupply() external pure returns (uint256);
+
+    function transferAll(address to) external returns (bool);
+
+    function transferAllFrom(address from, address to) external returns (bool);
+
+    // AMPL specific
+    function rebase(uint256 epoch, int256 supplyDelta) external returns (uint256);
```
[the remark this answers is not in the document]
nithinkrishna: Makes sense 👍  
We can have a seperate interface for admin only functions

nithinkrishna:
@brandoniles  Added the orchestrator too

contracts/UFragments.sol
```diff
@@ -144,6 +144,15 @@ contract UFragments is ERC20Detailed, Ownable {
         emit Transfer(address(0x0), owner_, _totalSupply);
     }
 
+    /*
+     * @notice A multi-chain AMPL interface method. On the base-chain contract it reflects
+     *         the value in the contract store.
+     * @return globalAMPLSupply The current total supply.
+     */
+    function globalAMPLSupply() external view returns (uint256) {
```
brandoniles, line 152: I'd prefer if the base AMPL version can exist as it is, without its interface being influenced by the need of satellite chains' amples. It also duplicates the totalSupply() function. Is this required by the bridge code?

Same with the policy.

contracts/interfaces/IAMPLGovernance.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity 0.4.24;
```
brandoniles, line 1: I think it's ok if the governance contracts simply declare the interface that's needed in the files where it's used. (Actually, I don't think they'd actually be used anyway since they'd be called in a generalized fashion from the voting contract). 

So I don't think we need this file.

contracts/interfaces/IAMPLSupplyPolicy.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity 0.4.24;
```
brandoniles, line 1: Similar to IAMPLGovernance, don't think we need this.

Oh wait, is this for the sattelite chains?

contracts/interfaces/IAmpleforth.sol
```diff
@@ -0,0 +1,25 @@
...
+interface IAmpleforth {
+    function rebase() external;
+
+    function epoch() external view returns (uint256);
+
+    function lastRebaseTimestampSec() external view returns (uint256);
+
+    function inRebaseWindow() external view returns (bool);
+
+    event LogRebase(
```
brandoniles, line 13: Does it make sense to include events in interaces? Only offchain services have access to logs, and they'd use an abi instead of a solidity interface.

contracts/interfaces/IAMPLSupplyPolicy.sol
nithinkrishna: No not for satellite chains

contracts/interfaces/IAmpleforth.sol
nithinkrishna: Makes sense 👍

contracts/UFragments.sol
nithinkrishna: Makes sense 👍  Removing this

brandoniles (approved):
Looks good. Curious to hear @thegostep 's take as well.

PR #195, opened 2021-02-03
Upgrading token compiler version

nithinkrishna:
The AMPL ERC-20 (0.4.24) is currently dependent on Openzeppelin's `openzeppelin-eth` (v2.0.2) which has now been rebranded and upgraded to [openzeppelin-contracts-upgradeable](http://github.com/openZeppelin/openzeppelin-contracts-upgradeable/). The current version of openzeppelin contracts (v3.3.0) run on a more recent solidity version (0.6.12) but have modified the storage layout. 

To get around this (safest & dirty solution): 
* [Copied all external dependency contracts locally](https://github.com/ampleforth/uFragments/pull/195/commits/0598c766151322d7145b69dc9a0dfb6458a1f355) to `contracts/_external`
* [Bumped up the solidity compiler version in dependency](https://github.com/ampleforth/uFragments/pull/195/commits/2f32adba45aead6983b62d0b965b20e164540614) for the `_external` contracts with necessary code changes 
* [Bumped up the solidity compiler version in all local contracts](https://github.com/ampleforth/uFragments/pull/195/commits/43316150a534dc88007731c29353a85573592795) 

I've updated all the contracts to 0.6.12 since that's the version the most recent openzeppelin contracts are using and I'm assuming that what the community agrees as the most stable. We can bump up to 0.7.x if we wanted to ..

### DIFF
--- a/contracts/_external/ERC20Detailed.sol
+++ b/contracts/_external/ERC20Detailed.sol
@@ -1,8 +1,7 @@
-pragma solidity ^0.4.24;
+pragma solidity 0.6.12;
 
-import "../../zos-lib/Initializable.sol";
+import "./Initializable.sol";
 import "./IERC20.sol";
-
 
 /**
  * @title ERC20Detailed token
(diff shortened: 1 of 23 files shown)


contracts/_external/ERC20Detailed.sol
```diff
@@ -0,0 +1,49 @@
...
+import "./Initializable.sol";
+import "./IERC20.sol";
+
+/**
+ * @title ERC20Detailed token
+ * @dev The decimals are only for visualization purposes.
+ * All the operations are done using the smallest and indivisible token unit,
+ * just as on Ethereum all the operations are done in wei.
+ */
+abstract contract ERC20Detailed is Initializable, IERC20 {
```
nithinkrishna, line 12: @thegostep  would this change the storage layout?

contracts/_external/ERC20Detailed.sol
```diff
@@ -1,46 +1,49 @@
...
-
-  /**
-   * @return the number of decimals of the token.
-   */
-  function decimals() public view returns(uint8) {
-    return _decimals;
-  }
-
-  uint256[50] private ______gap;
+abstract contract ERC20Detailed is Initializable, IERC20 {
```
nithinkrishna, line 12: @thegostep would this change the storage layout by any chance?

nithinkrishna:
The original dependencies are here: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/tree/b047d28476e274bca2a6091c280a7f55afe74e10

brandoniles (commented):
Can we also remove the references to openzeppelin-eth from package.json and yarn.lock?

thegostep:
I would bump to 0.7 unless there is a clear reason to stick to 0.6

contracts/_external/ERC20Detailed.sol
thegostep: unsure what this relates to

thegostep: in general - if the order of inheritance is unchanged and the order of storage variable declarations is unchanged, it should not modify storage layout

thegostep (approved):

contracts/_external/ERC20Detailed.sol
nithinkrishna: Makes sense. Was asking about making the contract `abstract`, but that shouldn't affect anything.

Issue #192, opened 2021-01-28
Orchestrator V2


nithinkrishna:
Contributors can pick up of the issues listed here (all are good first issues, great to get started):

- [x] The Orchestrator was created to disregard failed transactions. However it was [patched](https://github.com/ampleforth/uFragments/pull/153/) to revert if one of the listed transactions fail (to prevent a gas underprice attack). Update the [inline code documentation](https://github.com/ampleforth/uFragments/blob/master/contracts/Orchestrator.sol#L39) to reflect the patched change].

- [ ] Replace `externalCall` with `.call` 
- [ ] The orchestrator imports the entire `UFragmentsPolicy.sol`.  The typical way to address this is to just import the relevant interface, instead of importing the full contract. This helps reduce the deployed contract size.
- [ ] The current Orchestrator contract uses require(msg.sender == tx.origin) to prevent contracts from calling rebase. But tx.origin may be deprecated in the future. Implement one of these alternate mechanisms described [here](https://blog.openzeppelin.com/bypassing-smart-contract-timelocks).

- [ ] Use non upgradable version of `Ownable` in The Orchestrator contract.

tedwu13:
@nithinkrishna Can you explain a little more about `Replace externalCall with .call`? I see that externalCall function already has a .call function wrapped inside of the inline assembly.

nithinkrishna:
@tedwu13 With solidity 0.4.24 we had to use assembly to make external calls. With later versions you can use `address.call(data)` to make an external call

https://docs.soliditylang.org/en/v0.7.0/control-structures.html?highlight=call#external-function-calls

```
# something like?
Transaction storage t = transactions[i];
(bool result, bytes memory reason) = t.destination.call(t.data);
```

baribari1:
I think `tx.origin` is here to stay, in order to protect backwards compatibility, see below.
https://github.com/ethereum/solidity/issues/683

PR #203, opened 2021-02-18
Update inline documentation to document Orchestrator's revert

tedwu13:
## Overview
Based on one of the help wanted issues [here](https://github.com/ampleforth/uFragments/issues/192), I updated the inline documentation to reflect the [patched change](https://github.com/ampleforth/uFragments/pull/153/). 

Currently if any of the transactions failed, the rebase function will ignore the failed transaction and continue with other transactions. However with the added revert call, the Orchestrator will revert the state changes and return a `Transaction failed` message.

## Reviewers
@nithinkrishna


nithinkrishna (approved):

nithinkrishna:
Thanks @tedwu13!

PR #207, opened 2021-04-02
Orchestrator Updates

aalavandhan:
Fixes issues listed in #192

* The orchestrator imports the entire UFragmentsPolicy.sol. The typical way to address this is to just import the relevant interface, instead of importing the full contract. This helps reduce the deployed contract size.
* Use the revert message to indicate the index of the external call transaction which failed
* Use non upgradable version of Ownable in The Orchestrator contract.


nithinkrishna:
@tedwu13 you can review this too

contracts/_external/BytesLib.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Unlicensed
```
brandoniles, line 1: Is this audited or used in production anywhere?
Same with StringUtils

contracts/Orchestrator.sol
```diff
@@ -1,53 +1,72 @@
...
+                (bool success, bytes memory reason) = t.destination.call(t.data);
+
+                // Critical transaction failed, revert with message
+                if (!success && t.critical) {
+                    revert(buildRevertReason(i, reason));
+                }
+
+                // Non-Critical transaction failed, log error and continue
+                if (!success) {
+                    emit TransactionFailed(i);
```
brandoniles, line 69: I was thinking we could also include an error message in this log message. But I'm thinking maybe not, now. A non-critical TX might exist for awhile if we're waiting on an external system to fix itself. And we don't want to risk hitting gas limits unnecessarily.

Critical ones will have to get fixed fast or removed, so I think it's good to have there.

contracts/Orchestrator.sol
```diff
@@ -80,15 +105,68 @@ contract Orchestrator is Ownable {
...
+     */
+    function buildRevertReason(uint16 txIndex, bytes memory reason)
+        internal
+        pure
+        returns (string memory)
+    {
+        return
+            string(
+                abi.encodePacked(
+                    "Orchestrator: critical {job} reverted with {reason}:",
```
brandoniles, line 148:
```suggestion
                    "Orchestrator: critical index:{job} reverted with: {reason}:",
```

contracts/Orchestrator.sol
```diff
@@ -1,53 +1,72 @@
...
+    // events
+    event TransactionFailed(uint16 index);
 
     /**
      * @param policy_ Address of the UFragments policy.
      */
-    constructor(address policy_) public {
-        Ownable.initialize(msg.sender);
-        policy = UFragmentsPolicy(policy_);
+    constructor(address policy_) {
```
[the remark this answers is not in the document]
nithinkrishna: not really required with newer versions.
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol#L26

But will add it for the sake of being explicit

contracts/Orchestrator.sol
nithinkrishna: We could add it for consistency. Dont think we risk hitting gas limits anytime soon ..

contracts/_external/BytesLib.sol
nithinkrishna: The 2 external libraries
* [solidity-bytes-utils](https://github.com/GNSPS/solidity-bytes-utils) is used by Autherium, Optimism and a few others.

A recent audit found a minor issue with the `slice` method (which we use). But we should be using v0.1.2 which fixes this (just updated it)..

https://github.com/GNSPS/solidity-bytes-utils/blob/master/README.md

```
A bug regarding zero-length slices was disclosed by @MrChico following an audit to the Optimism codebase.

The exact bug happened under the following conditions: if memory slots higher then the current free-memory pointer
 were tainted before calling the slice method with a desired length of 0, the returned bytes array, instead of being a 
zero-length slice was an array of arbitrary length based on the values that previously populated that memory region.

Overall, the usage of zero-length slices should be pretty unusual and, as such, hopefully, this bug does not have far-
reaching implications. Nonetheless, please update the library to the new version if you're using it in production.

TL;DR: if you're using the slice method with a length parameter of '0' in your codebase, please update to version 0.1.2
 of the bytes library ASAP!
```

* There's aren't a widely used library for String utilities in solidity. The most used one is [Arachnid/solidity-stringutils](https://github.com/Arachnid/solidity-stringutils/blob/master/src/strings.sol). We could use that.

However since we dont require extensive string operations just a function to convert `uint` to `bytes/string`, I used [pipermerriam/ethereum-string-utils](https://github.com/pipermerriam/ethereum-string-utils/blob/master/contracts/StringLib.sol).

I dont think either of these libraries are audited we should do the due diligence ourselves.

***

The first library is not very critical as its invoked only when the transaction reverts. 
The 2nd one is probably a little more higher risk, but it's a very small function which I assume all of us can manuall verify.

contracts/Orchestrator.sol
```diff
@@ -1,53 +1,72 @@
-pragma solidity 0.7.6;
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.2;
```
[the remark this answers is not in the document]
nithinkrishna: I feel like keeping up with the latest version that [openzeppelin contracts](https://github.com/OpenZeppelin/openzeppelin-contracts) use is a good practice. 

@thegostep thoughts?

PR #223, opened 2021-12-20
Bumped up hardhat, ethers and oz contracts

aalavandhan:
Cleanup


brandoniles (commented):
Looks good

What motivated the change to oz-contracts-4.4.1 and the explicit inclusion of the version number though?

aalavandhan:
Seem to be running into dependency naming conflicts when importing 'ampleforth-contracts' as a dependency to 'cross-chain-ample' (which uses a different version of the oz contracts)

brandoniles (approved):
LGTM

PR #228, opened 2022-02-23
Market oracle

nms-7:
Migrates MedianOracle contract to this repository, also migrating the testing from truffle/js to hardhat/ts. 

Open questions: 
- Deployment - no migration file in the market-oracle repo, uncertainty around plans to actually deploy new oracle with any updated params or feature improvements
- There's a TODO left in the gas testing to fail if gas utilization changes – however it does change for some iterations
- Any new features to be discussed


package.json
```diff
@@ -30,7 +30,8 @@
     "lint"
   ],
   "dependencies": {
-    "openzeppelin-contracts-4.4.1": "https://github.com/OpenZeppelin/openzeppelin-contracts.git#4.4.1"
+    "openzeppelin-contracts-4.4.1": "https://github.com/OpenZeppelin/openzeppelin-contracts.git#4.4.1",
+    "openzeppelin-solidity": "2.0.0"
```
brandoniles, line 34: OpenZeppelin has retired the package openzeppelin-solidity for openzepelin-contracts. The other code in this repo already uses the newer package.

I bet we can drop-in replace the the import statements in the oracle code with the new package reference. This will keep our dependencies cleaner.

The one wrinkle might be the solidity version in the oracle. It could be that the new package isn't compatible with the older solidity compiler version.

Could you give it a shot?

nms-7: Agree this dirties the dependencies. Unfortunately, the versions of solidity are not compatible. 

Our other option is to additionally store the 0.4.24 (or compatible) helper contracts in the this repo, thoughts?

brandoniles: I'm good with it is, then. Thanks for checking!

brandoniles (commented):
LGTM, but please what for @ahnaguib or @aalavandhan 's approval before merging.

test/unit/MedianOracle_gas_cost.ts
```diff
@@ -0,0 +1,49 @@
+// TODO(naguib): Fail tests if gas utilization changes
+import { ethers } from 'hardhat'
+import { BigNumber, Contract, ContractFactory, Signer } from 'ethers'
```
aalavandhan, line 3: I think we can just get rid of this. We wrote this when there wasn't a lot of tooling for solidity development.

We now have a hardhat plugin built in which does this out of the box ,,
https://www.npmjs.com/package/hardhat-gas-reporter

contracts/MedianOracle.sol
```diff
@@ -0,0 +1,223 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
```
aalavandhan, line 3: I wonder if you can specify the version explicitly as i've done here .. 

https://github.com/ampleforth/cross-chain-ample/pull/60/files#diff-ff4eb0f76489225a40c20425b35dd4ae531717d0f53736363ad7b1bf62b681b6R4

aalavandhan:
Nice. Just a couple of comments ..

contracts/MedianOracle.sol
aalavandhan, line 3:
```suggestion
import {SafeMath} from "openzeppelin-solidity/contracts/math/SafeMath.sol";
```

contracts/MedianOracle.sol
```diff
@@ -0,0 +1,223 @@
...
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
```
aalavandhan, line 4:
```suggestion
import {Ownable} from "openzeppelin-solidity/contracts/ownership/Ownable.sol";
```

aalavandhan:
I wonder if we can just use the latest solidity versions (and oz dependencies) for the market oracle contracts you've imported.  If we redeploy we'll bump them up anyway.  

We could add an issue and get back to it in a later PR, no rush.

contracts/MedianOracle.sol
nms-7: so I tried a few things here. I could `yarn add` from the github path as you had, in the case "https://github.com/OpenZeppelin/openzeppelin-**solidity**.git#2.0.0" actually resolves to "https://github.com/OpenZeppelin/openzeppelin-**contracts**.git#2.0.0", but the package name remains "openzeppelin-solidity" for purposes of importing from the contract code

nms-7:
> 

Open to doing this. Of course, would not match the existing deployed oracle, so maybe better for a separate PR, possibly including any feature update proposals

aalavandhan (approved):

PR #253, opened 2023-05-04
Orchestrator whitelist

nms-7:
adds whitelist to orchestrator contract so that whitelisted contracts can call rebase

### DIFF
--- a/contracts/Orchestrator.sol
+++ b/contracts/Orchestrator.sol
@@ -22,6 +22,8 @@ contract Orchestrator is Ownable {
     // Stable ordering is not guaranteed.
     Transaction[] public transactions;
 
+	address[] private whitelist;
+
     IUFragmentsPolicy public policy;
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/Orchestrator.sol
```diff
@@ -22,6 +22,8 @@ contract Orchestrator is Ownable {
     // Stable ordering is not guaranteed.
     Transaction[] public transactions;
 
+    address[] private whitelist;
```
aalavandhan, line 25: Maybe use `EnumerableSet` here, so we don't have to loop through the list to check?

contracts/Orchestrator.sol
```diff
@@ -42,7 +44,24 @@ contract Orchestrator is Ownable {
      */
     function rebase() external {
         require(msg.sender == tx.origin); // solhint-disable-line avoid-tx-origin
```
aalavandhan, line 46: We could just have one `rebase` function but add a modifier where we check if  either `msg.sender == tx.origin` or `msg.sender` is part of the whitelist

aalavandhan:
We could also update to the latest solidity version as we do this

PR #265, opened 2023-10-04
Target rate update

aalavandhan:
* Policy now gets the cpi adjusted target rate as an output from the cpi oracle (rather than getting the new CPI and scaling it to then compute the current target).
* Median oracle now has an additional scaling factor parameter (settable by the owner). The final `output = median*scalar`


contracts/UFragmentsPolicy.sol
```diff
@@ -29,6 +29,7 @@ contract UFragmentsPolicy is Ownable {
     using SafeMathInt for int256;
     using UInt256Lib for uint256;
 
+    /// @notice DEPRECATED.
     event LogRebase(
```
aalavandhan, line 33: We need to update and redeploy the subgraph index based on either `LogRebase` and `Rebase`.

contracts/UFragmentsPolicy.sol
```diff
@@ -37,16 +38,27 @@ contract UFragmentsPolicy is Ownable {
         uint256 timestampSec
     );
 
+    event Rebase(
```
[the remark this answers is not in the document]
aalavandhan: > Also worth checking if events occupy any storage in the deployed code and hence need to be added at the end for upgrades?

Adding events shouldn't modify the storage layout. Fwiw, The oz tools check for conflicts and I didn't see any.

contracts/MedianOracle.sol
```diff
@@ -47,6 +50,9 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
     // This is needed so that timestamp of 1 is always considered expired.
     uint256 private constant MAX_REPORT_EXPIRATION_TIME = 520 weeks;
 
+    // The final mediam value is scaled by dividing by this factor.
```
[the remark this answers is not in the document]
aalavandhan: Agreed 👍

contracts/UFragmentsPolicy.sol
```diff
@@ -37,16 +38,27 @@ contract UFragmentsPolicy is Ownable {
         uint256 timestampSec
     );
 
+    event LogRebaseV2(
+        uint256 indexed epoch,
+        uint256 exchangeRate,
+        uint256 targetRate,
+        int256 requestedSupplyAdjustment,
+        uint256 timestampSec
```
brandoniles, line 46: In theory, the timestamp is available in the block header, but I can see how it might be easier for downstream consumers to have this natively in the log stream with the other data.

Is subgraph able to index block timestamps?

contracts/UFragmentsPolicy.sol
```diff
@@ -37,16 +38,27 @@ contract UFragmentsPolicy is Ownable {
...
-    // Provides the current CPI, as an 18 decimal fixed point number.
+    // Provides the cpi adjusted price target, as an 18 decimal fixed point number.
     IOracle public cpiOracle;
 
     // Market oracle provides the token/USD exchange rate as an 18 decimal fixed point number.
     // (eg) An oracle value of 1.5e18 it would mean 1 Ample is trading for $1.50.
     IOracle public marketOracle;
 
-    // CPI value at the time of launch, as an 18 decimal fixed point number.
+    // @notice DEPRECATED.
```
brandoniles, line 58: You could also rename the variable itself to something like `baseCpi_DEPRECATED`. This shouldn't impact storage layout, and it makes it clear if future code accidentally references it.

contracts/MedianOracle.sol
```diff
@@ -47,6 +50,9 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
     // This is needed so that timestamp of 1 is always considered expired.
     uint256 private constant MAX_REPORT_EXPIRATION_TIME = 520 weeks;
 
+    // The final mediam value is scaled by multiplying by this factor.
```
brandoniles, line 53:
```suggestion
    // The final median value is scaled by multiplying by this factor.
```

contracts/MedianOracle.sol
```diff
@@ -182,7 +197,9 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
             return (0, false);
         }
 
-        return (Select.computeMedian(validReports, size), true);
+        uint256 result = Select.computeMedian(validReports, size);
+        result = result.mulDiv(scalar, 10**18);
```
brandoniles, line 201: This line enforces that `scalar` is in 18 decimals format, but that's not documented.

Let's update the `scalar` NatSpec to include this.
I'd be ok with also making an immutable `SCALAR_DECIMALS` property for the contract, which shouldn't use any extra storage.

contracts/MedianOracle.sol
```diff
@@ -67,6 +73,7 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
         reportExpirationTimeSec = reportExpirationTimeSec_;
         reportDelaySec = reportDelaySec_;
         minimumProviders = minimumProviders_;
+        scalar = 10**18;
```
brandoniles, line 76: Should we also add a scalar param to the `init()` inputs?

Wouldn't affect our deployment, but I'd expect future deployments to consider initing it to something.

contracts/UFragmentsPolicy.sol
aalavandhan: I was reluctant to run an upgrade with `--force` but seems like they now allow a clean way to rename variables. 
https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/531

contracts/MedianOracle.sol
aalavandhan: All numbers in this contract use the same 18 digit decimal representation so we can define DECIMALS=18, rather than SCALAR_DECIMALS=18 and DATA_DECIMALS=18

contracts/MedianOracle.sol
aalavandhan: Sounds good

contracts/UFragmentsPolicy.sol
aalavandhan: yea we can get rid of the timestamp.

nms-7 (approved):

contracts/MedianOracle.sol
```diff
@@ -116,7 +120,7 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
 
     /**
      * @notice Pushes a report for the calling provider.
-     * @param payload is expected to be 18 decimal fixed point number.
+     * @param payload is expected to be a fixed point number with 18 decimals.
```
brandoniles, line 123:
```suggestion
     * @param payload is expected to be a fixed point number with `DECIMALS` decimals.
```

Is this worth it or just more confusing?
https://docs.soliditylang.org/en/latest/natspec-format.html#dynamic-expressions

brandoniles (approved):
LGTM

contracts/MedianOracle.sol
```diff
@@ -56,17 +64,20 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
...
      * @param minimumProviders_ The minimum number of providers with valid
      *                          reports to consider the aggregate report valid.
+     * @param scalar_ The scaling factor applied on the result.
      */
     function init(
         uint256 reportExpirationTimeSec_,
         uint256 reportDelaySec_,
-        uint256 minimumProviders_
+        uint256 minimumProviders_,
+        uint256 scalar_
```
ahnaguib, line 73: Thinking about this more I think we need different scalar for different providers when using different sources for target price (we can do that in a later change though)

ahnaguib (approved):

PR #281, opened 2024-08-21
Remove emit ReportTimestampOutOfRange so getData is read only

ahnaguib:
Remove emit ReportTimestampOutOfRange so getData is read only

### DIFF
--- a/contracts/MedianOracle.sol
+++ b/contracts/MedianOracle.sol
@@ -34,7 +34,7 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
 
     event ProviderAdded(address provider);
     event ProviderRemoved(address provider);
-    // Deprecated: event ReportTimestampOutOfRange(address provider);
+    event ReportTimestampOutOfRange(address provider);
     event ProviderReportPushed(address indexed provider, uint256 payload, uint256 timestamp);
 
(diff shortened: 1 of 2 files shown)


aalavandhan (approved):

aalavandhan:
Yea. Good call. Was meaning to remove this.

contracts/MedianOracle.sol
```diff
@@ -186,10 +186,10 @@ contract MedianOracle is OwnableUpgradeable, IOracle {
                     .timestamp;
                 if (reportTimestampPast < minValidTimestamp) {
                     // Past report is too old.
-                    emit ReportTimestampOutOfRange(providerAddress);
+                    // Deprecated: emit ReportTimestampOutOfRange(providerAddress);
```
aalavandhan, line 189: I think we can just remove it, rather than commenting it out. Dont think any of our off-chain services rely on this log.

aalavandhan (approved):

contracts/MedianOracle.sol
ahnaguib: I think it is a good idea to leave so that previous logs are still understandable

brandoniles (commented):
Looks good... can we also add the `view` modifier to the `getData()` signature?

PR #287, opened 2024-10-09
Policy interface update

aalavandhan:
Added cpi and exchange rate getters


nms-7 (approved):

brandoniles (commented):
I think we should add `getTargetRate()` to the IAmpleforth interface.

But don't think we need `getExchangeRate()`

PR #290, opened 2024-10-17
Updated invalid oracle input handling

aalavandhan:
We rebase with zero supply change when oracle inputs are invalid at the time of rebase


brandoniles (approved):
I suppose there is a case where the oracle report becomes valid later inside the same rebase window. In that case, an otherwise successful non-zero rebase would be blocked by the zero rebase that came before it.

But I don't see any good way to guard against that and still get this functionality.

ahnaguib:
LGTM!

PR #294, opened 2024-10-29
Rebase Circuit Breaker

aalavandhan:
Owner can limit the total supply contraction over an x-epoch period

### DIFF
--- a/contracts/UFragmentsPolicy.sol
+++ b/contracts/UFragmentsPolicy.sol
@@ -236,7 +236,7 @@ contract UFragmentsPolicy is Ownable {
     /**
      * @notice Sets the parameters which control rebase circuit breaker.
      * @param epochLookback_ The number of rebase epochs to look back.
-     * @param tolerableDeclinePercentage_ The maximum supply decline percentage which is allowed
+     * @param tolerableDeclinePercentage_ The supply decline percentage which is allowed
      *        within the defined look back period.
      */
     function setRebaseCircuitBreakerParameters(
@@ -377,19 +377,15 @@ contract UFragmentsPolicy is Ownable {
 
         // When supply is decreasing:
         // We limit the supply delta, based on recent supply history.
-        if (rebasePercentage < 0) {
-            int256 maxSupply = currentSupply;
-            for (
-                uint256 e = ((epoch > epochLookback) ? (epoch - epochLookback) : 0);
-                e < epoch;
-                e++
-            ) {
-                int256 epochSupply = supplyHistory[e].toInt256Safe();
-                if (epochSupply > maxSupply) {
-                    maxSupply = epochSupply;
-                }
+        if (rebasePercentage < 0 && epoch > epochLookback) {
+            int256 allowedMin = supplyHistory[epoch - epochLookback]
+                .toInt256Safe()
+                .mul(ONE.sub(tolerableDeclinePercentage))
+                .div(ONE);
+            if (allowedMin > currentSupply) {
+                // NOTE: Allowed minimum supply can only be at most the current supply.
+                allowedMin = currentSupply;
             }
-            int256 allowedMin = maxSupply.mul(ONE.sub(tolerableDeclinePercentage)).div(ONE);
             if (newSupply < allowedMin) {
                 newSupply = allowedMin;
             }
--- a/test/unit/UFragmentsPolicy.ts
+++ b/test/unit/UFragmentsPolicy.ts
@@ -1409,7 +1409,21 @@ describe('UFragmentsPolicy', function () {
         expect(
           (await parseRebaseEvent(uFragmentsPolicy.rebase()))
             .requestedSupplyAdjustment,
-        ).to.eq(-49)
+        ).to.eq(-51)
+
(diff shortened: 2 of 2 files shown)


contracts/UFragmentsPolicy.sol
```diff
@@ -340,7 +371,28 @@ contract UFragmentsPolicy is Ownable {
...
-        return uFrags.totalSupply().toInt256Safe().mul(rebasePercentage).div(ONE);
+
+        int256 currentSupply = uFrags.totalSupply().toInt256Safe(); // (or) supplyHistory[epoch]
+        int256 newSupply = ONE.add(rebasePercentage).mul(currentSupply).div(ONE);
+
+        // When supply is decreasing:
+        // We limit the supply delta, based on recent supply history.
+        if (rebasePercentage < 0) {
+            int256 maxSupplyInHistory = currentSupply;
+            for (uint8 i = 1; i < epochLookback && epoch > i; i++) {
```
brandoniles, line 382: This structure seems much easier to read and harder to make mistakes with. Uses epochs directly, rather than offsets.
```suggestion
            for (uint256 e = epoch - epochLookback; e < epoch; e++) {
```

contracts/UFragmentsPolicy.sol
```diff
@@ -340,7 +371,28 @@ contract UFragmentsPolicy is Ownable {
...
+                int256 epochSupply = supplyHistory[epoch - i].toInt256Safe();
+                if (epochSupply > maxSupplyInHistory) {
+                    maxSupplyInHistory = epochSupply;
+                }
+            }
+            int256 allowedSupplyMinimum = maxSupplyInHistory
+                .mul(ONE.sub(tolerableDeclinePercentage))
+                .div(ONE);
+            newSupply = (newSupply > allowedSupplyMinimum) ? newSupply : allowedSupplyMinimum;
```
brandoniles, line 391: More a fan of this style that avoids no-op assignments:
```
            if (newSupply < allowed) {
                newSupply = allowed;
            }
```

contracts/UFragmentsPolicy.sol
```diff
@@ -340,7 +371,28 @@ contract UFragmentsPolicy is Ownable {
...
+            require(newSupply <= currentSupply);
```
brandoniles, line 392: This could cause the rebase operation to fail, which we've recently started to avoid.
Should we make this recover into a no-op supply change?

brandoniles (commented):
I was wondering if we should use a circular array, rather than a linearly increasing hashmap. The flexibility of the map the gas benefits, though, I agree. It could still be moved to in the future if it became a concern

contracts/UFragmentsPolicy.sol
aalavandhan: We can get rid of this require. If the implementation is correct, `newSupply <= currentSupply` should always be true. Given that supply is reducing, the limiting logic should never increase the supply.

contracts/UFragmentsPolicy.sol
aalavandhan: Runs into an error when epoch < lookback

contracts/UFragmentsPolicy.sol
```diff
@@ -340,7 +371,27 @@ contract UFragmentsPolicy is Ownable {
             rebaseFunctionUpperPercentage,
             rebaseFunctionGrowth
         );
-        return uFrags.totalSupply().toInt256Safe().mul(rebasePercentage).div(ONE);
+
+        int256 currentSupply = uFrags.totalSupply().toInt256Safe(); // (or) supplyHistory[epoch]
```
brandoniles, line 375:
```suggestion
        int256 currentSupply = uFrags.totalSupply().toInt256Safe();
```
I don't think the supplyHistory has been written to yet, in the core rebase flow.

contracts/UFragmentsPolicy.sol
```diff
@@ -340,7 +371,27 @@ contract UFragmentsPolicy is Ownable {
...
+        int256 newSupply = ONE.add(rebasePercentage).mul(currentSupply).div(ONE);
```
brandoniles, line 376: Looking for it, but I don't see this change... Where did rebasePercentage move from being 1-centered to 0-centered? (We add one here but not in the previous version's line)

contracts/UFragmentsPolicy.sol
```diff
@@ -340,7 +371,27 @@ contract UFragmentsPolicy is Ownable {
...
+
+        // When supply is decreasing:
+        // We limit the supply delta, based on recent supply history.
+        if (rebasePercentage < 0 && epoch > epochLookback) {
```
brandoniles, line 380: In the case that the supply history hasn't been written to yet, and is 0, then the below code basically no-ops? That's what it seems like. Do we have an explicit test case for this?

i.e.
1) Test for supply history lookup returns 0
2) Test for when epochLookback equals 0 (should basically be same as above since history will return 0)

brandoniles (commented):
I like the idea of having an owner-callable history cleanup function. I imagine it would take a start epoch, end epoch, and wouldn't zero-out anything within the existing lookback window.

PR #297, opened 2025-03-25
Refactor rebase function parameters to separate positive and negative growth

ahnaguib:


contracts/UFragmentsPolicy.sol
```diff
@@ -94,9 +94,10 @@ contract UFragmentsPolicy is Ownable {
 
     // DECIMALS decimal fixed point numbers.
     // Used in computation of  (Upper-Lower)/(1-(Upper/Lower)/2^(Growth*delta))) + Lower
-    int256 public rebaseFunctionLowerPercentage;
-    int256 public rebaseFunctionUpperPercentage;
-    int256 public rebaseFunctionGrowth;
+    int256 public rebaseFunctionNegativePercentageLimit;
```
aalavandhan, line 97: Here we are basically renaming 3 storage variables and adding one more. Changes look upgrade safe. 👍 

We can use the `/// @custom:oz-renamed-from operator` natspec comment to keep track of pervious names. It helps the oz upgrade tooling keep track of the storage state ..

https://github.com/ampleforth/ampleforth-contracts/blob/master/.openzeppelin/mainnet.json#L155

https://forum.openzeppelin.com/t/rename-struct-property/38519/4

ahnaguib: Thanks!

PR #311, opened 2026-06-02
DexOracle: indirect AMPL/USDC 24h TWAP provider

ahnaguib:
No direct AMPL/USDC UniswapV2 market exists on mainnet, so this oracle bridges through WETH: AMPL/USDC = (AMPL/WETH) x (WETH/USDC), reporting an 18-decimal TWAP to a MedianOracle on the 24h rebase cadence. update() opens the measurement window right after rebase (via the Orchestrator tx list, gated to the daily update window) and pushReport() closes it ~2h before the next rebase (gated by minReportTimeIntervalSec) so the report ages past the MedianOracle security window.

### DIFF
--- a/contracts/DexOracle.sol
+++ b/contracts/DexOracle.sol
@@ -1,0 +1,336 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.4;
+
+import {Ownable} from "./_external/Ownable.sol";
+import {SafeMath} from "./_external/SafeMath.sol";
+import {IUniswapV2Pair} from "./_external/IUniswapV2Pair.sol";
+import {UniswapV2OracleLibrary} from "./_external/UniswapV2OracleLibrary.sol";
+
+interface IERC20Decimals {
+    function decimals() external view returns (uint8);
+}
+
+interface IMedianOracle {
+    function pushReport(uint256 payload) external;
+}
+
+/**
+ * @title DexOracle
+ *
+ * @notice Computes a 24h time-weighted average price (TWAP) for an asset pair
+ *         that has no direct UniswapV2 market by chaining two underlying
+ *         markets, and reports the result to a MedianOracle instance.
+ *
+ *         For AMPL/USDC the price is bridged through WETH:
+ *
+ *             AMPL/USDC = (AMPL/WETH) * (WETH/USDC)
+ *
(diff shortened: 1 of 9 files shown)


contracts/DexOracle.sol
```diff
@@ -0,0 +1,302 @@
...
+        require(blockTimestampLast > 0, "DexOracle: UPDATE_NEVER_CALLED");
+
+        (
+            uint256 leg1Cumulative,
+            uint256 leg2Cumulative,
+            uint32 blockTimestamp
+        ) = _currentCumulatives();
+        unchecked {
+            // Wraparound is desired; both timestamps are taken mod 2**32.
+            timeElapsed = blockTimestamp - blockTimestampLast;
```
brandoniles, line 267: I don't see why this operation should be unchecked...

But it seems like the subtractions in the computatio of priceLeg1 and priceLeg2 should be?

contracts/DexOracle.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * @notice Opens a fresh measurement window by snapshotting the current
+     *         price cumulatives. Intended to be appended to the Orchestrator's
+     *         transaction list so it runs immediately after each rebase.
+     * @dev Restricted to the Orchestrator and the owner, who are trusted to
+     *      open the window on schedule; both may call at any time.
+     */
+    function update() external {
+        require(msg.sender == orchestrator || isOwner(), "DexOracle: UNAUTHORIZED");
```
ahnaguib: allow calls from anyone if last checkpoint is stale

contracts/DexOracle.sol
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * @notice Computes the chained TWAP over the current measurement window and
+     *         reports it to the MedianOracle. Intended to be called ~2h before
+     *         the next rebase, leaving the report to age past the MedianOracle
+     *         report-delay window before it is consumed.
+     * @return price The reported AMPL/USDC price as an OUTPUT_DECIMALS number.
+     */
+    function pushReport() external returns (uint256 price) {
+        uint32 timeElapsed;
+        (price, timeElapsed) = _computePrice();
```
ahnaguib, line 198: add safeguards on pool size.

contracts/DexOracle.sol
ahnaguib: because uint32 will overflow in the year 2106

